PR #69, opened 2024-12-19
Refactor generators

### LINKED ISSUE #68: Replace deprecated generators with current ones
Some generators are deprecated. Let's switch to the latest ones. Sample both the current and updated generators and make sure their outputs match closely.

BowTiedRadone:
This PR refactors `fast-check` generators. A custom hex string generator, `hexaString`, has been added to replace the deprecated `hexaString` generator from `fast-check`. Clarity `buff` type generation now doubles the hex
length. Minor updates have been made to the `string-ascii` generator to ensure proper handling, including support for empty strings. Tests have been added to validate these changes and ensure correctness. Resolves #68.

Related references:
1. https://github.com/dubzzz/fast-check/commit/3f4f1203a8863c07d22b45591bf0de1fac02b948
3. https://github.com/hirosystems/stacks.js/blob/fd0bf26b5f29fc3c1bf79581d0ad9b89f0d7f15a/packages/common/src/utils.ts#L522


sample.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @param generatorB The second generator to compare.
+ * @param numSamples The number of samples to compare.
+ * @returns An array of result objects containing the sample number, values
+ * generated by each generator, and the match object containing whether the
+ * values match and the mismatched items.
+ */
+export const compareGenerators = (
+  generatorA: fc.Arbitrary<any>,
+  generatorB: fc.Arbitrary<any>,
+  numSamples = 10
```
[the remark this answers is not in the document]
moodmosaic: Then I'd rather hardcode `10` inside the function. Or even better, have `fc.nat` prodive a range between 7-20 for the number of samples.

sample.tests.ts
```diff
@@ -0,0 +1,65 @@
...
+        );
+        const isMatching = comparisonResults.every(
+          (result) => result.match.matches
+        );
+        const mismatchedItems = comparisonResults.flatMap(
+          (result) => result.match.mismatchedItems
+        );
+
+        expect(mismatchedItems).toEqual([]);
+        expect(isMatching).toBe(true);
```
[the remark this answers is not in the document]
moodmosaic: The `matches` can be inferred from the `mismatchedItems` then?

sample.ts
```diff
@@ -0,0 +1,52 @@
...
+ * values match and the mismatched items.
+ */
+export const compareGenerators = (
+  generatorA: fc.Arbitrary<any>,
+  generatorB: fc.Arbitrary<any>,
+  numSamples: number
+) => {
+  const results = [];
+  for (let i = 0; i < numSamples; i++) {
+    const seed = 123 * i;
```
moodmosaic, line 19: Since this test suite now uses fast-check, pass the seed as a parameter (same we did for the number of sample).

Then we can always set the global seed with fc.configureGlobal in the test suite root for replays.

sample.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @param numSamples The number of samples to compare.
+ * @returns An array of result objects containing the sample number, values
+ * generated by each generator, and the match object containing whether the
+ * values match and the mismatched items.
+ */
+export const compareGenerators = (
+  generatorA: fc.Arbitrary<any>,
+  generatorB: fc.Arbitrary<any>,
+  numSamples: number
+) => {
```
moodmosaic, line 16: It'd be nice to see the return type here.

sample.ts
```diff
@@ -0,0 +1,52 @@
...
+  const results = [];
+  for (let i = 0; i < numSamples; i++) {
+    const seed = 123 * i;
+
+    const valueA = fc.sample(generatorA, { seed: seed });
+    const valueB = fc.sample(generatorB, { seed: seed });
```
moodmosaic, line 22: Why we even need `compareGenerators` and `getMismatchedItems`. I'd delete the `sample.ts` file.

---

What I'd do instead in the actual tests we have is just

```ts
const valueA = fc.sample(generatorA, { seed: seed });
const valueB = fc.sample(generatorB, { seed: seed });
```

and compare the two values in the simplest possible way, e.g. JSON.stringify.

No need for loops:
- Just one run of fc.sample on both generators is enough.
- If the test is [erratic](https://wubw.github.io/2017/xunit_test_patterns/#erratic-test) we'll catch it sooner or later running the test-suite (as part of the CI or our everyday workflow).

moodmosaic (changes requested):

sample.ts
moodmosaic: Turns out this whole file is not needed at all. See https://github.com/stacks-network/rendezvous/pull/69#discussion_r1894582703.

sample.ts
moodmosaic: Ignore as per https://github.com/stacks-network/rendezvous/pull/69#discussion_r1894582703.

moodmosaic (approved):
LGTM

PR #79, opened 2025-01-02
Fix error message for contract not found in manifest

BowTiedRadone:
This PR fixes the error message when the specified contract cannot be found in the manifest file.

### DIFF
--- a/citizen.ts
+++ b/citizen.ts
@@ -227,7 +227,7 @@ export const buildRendezvousData = (
     };
   } catch (e: any) {
     throw new Error(
-      `Error processing contract ${contractName.split(".")[1]}: ${e.message}`
+      `Error processing "${contractName}" contract: ${e.message}`
     );
   }
 };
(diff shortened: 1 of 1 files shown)


citizen.ts
```diff
@@ -227,7 +227,7 @@ export const buildRendezvousData = (
     };
   } catch (e: any) {
     throw new Error(
-      `Error processing contract ${contractName.split(".")[1]}: ${e.message}`
+      `Error processing "${contractName}" contract: ${e.message}`
```
moodmosaic, line 230: Nice 👍

citizen.ts
```diff
@@ -256,7 +256,7 @@ const getSimnetPlanContractSource = (
 
   if (contractInfo == undefined) {
     throw new Error(
-      `Contract ${sutContractName} not found in the Clarinet.toml.`
+      `The "${sutContractName}" contract was not found in the Clarinet.toml.`
```
moodmosaic, line 259: In Unix, errors like ENOENT use the pattern “File not found.”

Imagine if it said, “The file not found” instead. What’s the advantage? Feels less natural (to me).

If consistency matters here, I’d stick with the convention.

BowTiedRadone: Or do you mean to remove "the" and "was"? If that's the case I agree.

moodmosaic: Yes

PR #81, opened 2025-01-06
Add burnchain advancement feature

### LINKED ISSUE #76: Enable Mining of Empty Bitcoin Blocks Between Calls
After Nakamoto, public function calls performed using `clarinet-sdk`, result in mining a new Stacks block. However, the burnchain remains untouched. This behavior leads to more blocks within a single tenure, rather than creating additional tenures.

To better simulate real-world scenarios, Rendezvous should support mining an empty number of Bitcoin blocks between calls. This would allow for:
- More accurate modeling of tenures in Stacks.
- Improved testing for scenarios dependent on burnchain progression.

This feature would help developers working on smart contracts that rely on the burnchain.

BowTiedRadone:
This PR introduces a seed-based method for advancing the burnchain in both invariant and property testing. Logs now display the current Stacks and Bitcoin block heights.

**Sample Output:**
```
# type=test

₿      1 Ӿ      6   wallet_4 [PASS] reverse test-reverse-string 
₿      1 Ӿ      7   wallet_7 [PASS] reverse test-reverse [1764733795]
₿      2 Ӿ      9   wallet_7 [PASS] reverse test-reverse-string toLoc
₿      2 Ӿ     10   wallet_3 [PASS] reverse test-reverse-string 7H0/t
₿      3 Ӿ     12   wallet_5 [PASS] reverse test-reverse-string -
```
```
# type=invariant

₿     48 Ӿ     95   deployer [PASS] counter invariant-counter-gt-zero 
₿     48 Ӿ     97   wallet_5        counter decrement 
₿     48 Ӿ     97   wallet_5 [PASS] counter invariant-counter-gt-zero 
₿     57 Ӿ    108   wallet_5        counter increment 
₿     57 Ӿ    108   wallet_8 [PASS] counter invariant-counter-gt-zero 
₿     57 Ӿ    110   deployer        counter add 2147483639
₿     57 Ӿ    110   wallet_4 [PASS] counter invariant-counter-gt-zero 
₿     57 Ӿ    112   wallet_2        counter increment 
₿     57 Ӿ    112   wallet_7 [PASS] counter invariant-counter-gt-zero 
₿     62 Ӿ    119   wallet_3        counter add 6
₿     62 Ӿ    119   wallet_7 [PASS] counter invariant-counter-gt-zero
```

Resolves #76.

### DIFF
--- a/invariant.ts
+++ b/invariant.ts
@@ -42,7 +42,7 @@ export const checkInvariants = (
 
   radio.emit(
     "logMessage",
-    `\nStarting invariant testing type for the ${sutContractName} contract...`
+    `\nStarting invariant testing type for the ${sutContractName} contract...\n`
   );
 
   const simnetAccounts = simnet.getAccounts();
@@ -64,6 +64,7 @@ export const checkInvariants = (
           rendezvousContractId: fc.constantFrom(...rendezvousList),
           sutCaller: fc.constantFrom(...eligibleAccounts.entries()),
           invariantCaller: fc.constantFrom(...eligibleAccounts.entries()),
+          mineBurnBlocks: fc.boolean(),
         })
         .chain((r) => {
           const functions = getFunctionsListForContract(
@@ -122,7 +123,16 @@ export const checkInvariants = (
               invariantArgsArb: fc.tuple(...invariantArgsArb),
             })
             .map((args) => ({ ...r, ...args }));
-        }),
+        })
+        .chain((r) =>
+          fc
+            .record({
+              burnBlocksToMine: r.mineBurnBlocks
+                ? fc.integer({ min: 1, max: 10 })
+                : fc.constant(0),
+            })
+            .map((burnBlocksToMine) => ({ ...r, ...burnBlocksToMine }))
+        ),
       (r) => {
         const selectedFunctionArgs = argsToCV(
           r.selectedFunction,
@@ -174,17 +184,25 @@ export const checkInvariants = (
 
             radio.emit(
               "logMessage",
-              `${dim(sutCallerWallet)}        ${getContractNameFromRendezvousId(
-                r.rendezvousContractId
-              )} ${underline(r.selectedFunction.name)} ${printedFunctionArgs}`
+              `₿ ${simnet.burnBlockHeight.toString().padStart(6)} ` +
+                `Ӿ ${simnet.blockHeight.toString().padStart(6)}   ` +
+                dim(`${sutCallerWallet}        `) +
+                `${getContractNameFromRendezvousId(r.rendezvousContractId)} ` +
+                `${underline(r.selectedFunction.name)} ` +
+                printedFunctionArgs
             );
           } else {
             radio.emit(
               "logMessage",
               dim(
-                `${sutCallerWallet}        ${getContractNameFromRendezvousId(
-                  r.rendezvousContractId
-                )} ${underline(r.selectedFunction.name)} ${printedFunctionArgs}`
+                `₿ ${simnet.burnBlockHeight.toString().padStart(6)} ` +
+                  `Ӿ ${simnet.blockHeight.toString().padStart(6)}   ` +
+                  `${sutCallerWallet}        ` +
+                  `${getContractNameFromRendezvousId(
+                    r.rendezvousContractId
+                  )} ` +
+                  `${underline(r.selectedFunction.name)} ` +
+                  printedFunctionArgs
               )
             );
           }
@@ -195,9 +213,12 @@ export const checkInvariants = (
           radio.emit(
             "logMessage",
             dim(
-              `${sutCallerWallet}        ${getContractNameFromRendezvousId(
-                r.rendezvousContractId
-              )} ${underline(r.selectedFunction.name)} ${printedFunctionArgs}`
+              `₿ ${simnet.burnBlockHeight.toString().padStart(6)} ` +
+                `Ӿ ${simnet.blockHeight.toString().padStart(6)}   ` +
+                `${sutCallerWallet}        ` +
+                `${getContractNameFromRendezvousId(r.rendezvousContractId)} ` +
+                `${underline(r.selectedFunction.name)} ` +
+                printedFunctionArgs
             )
           );
         }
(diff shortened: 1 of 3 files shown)


invariant.ts
```diff
@@ -42,7 +42,7 @@ export const checkInvariants = (
 
   radio.emit(
     "logMessage",
-    `\nStarting invariant testing type for the ${sutContractName} contract...`
+    `\nStarting invariant testing type for the ${sutContractName} contract...\n`
```
moodmosaic, line 45: Do we really need a new line after?

invariant.ts
```diff
@@ -64,6 +64,7 @@ export const checkInvariants = (
           rendezvousContractId: fc.constantFrom(...rendezvousList),
           sutCaller: fc.constantFrom(...eligibleAccounts.entries()),
           invariantCaller: fc.constantFrom(...eligibleAccounts.entries()),
+          mineBurnBlocks: fc.boolean(),
```
moodmosaic, line 67: This name doesn't clearly follow the "should" or "can" convention for booleans. Perhaps `canMineBlocks`?

invariant.ts
```diff
@@ -122,7 +123,16 @@ export const checkInvariants = (
               invariantArgsArb: fc.tuple(...invariantArgsArb),
             })
             .map((args) => ({ ...r, ...args }));
-        }),
+        })
+        .chain((r) =>
+          fc
+            .record({
+              burnBlocksToMine: r.mineBurnBlocks
+                ? fc.integer({ min: 1, max: 10 })
```
moodmosaic, line 131: Can we have `max` increase as the number of runs increases? Also, would that make sense? I think `10` might be too small for larger runs.

invariant.ts
```diff
@@ -174,17 +184,25 @@ export const checkInvariants = (
 
             radio.emit(
               "logMessage",
-              `${dim(sutCallerWallet)}        ${getContractNameFromRendezvousId(
-                r.rendezvousContractId
-              )} ${underline(r.selectedFunction.name)} ${printedFunctionArgs}`
+              `₿ ${simnet.burnBlockHeight.toString().padStart(6)} ` +
```
moodmosaic, line 187: Is `6` space enough for say 1M runs?

invariant.ts
```diff
@@ -254,15 +277,27 @@ export const checkInvariants = (
...
             )
           );
 
           // Re-throw the error for fast-check to catch and process.
           throw error;
         }
+
+        try {
+          if (r.mineBurnBlocks) {
+            simnet.mineEmptyBurnBlocks(r.burnBlocksToMine);
```
moodmosaic, line 296: Perhaps just `r.burnBlocks` would be easier, cleaner, and more action-oriented, suggesting blocks that are meant to be burned.

invariant.ts
```diff
@@ -254,15 +277,27 @@ export const checkInvariants = (
...
+          }
+        } catch (error: any) {
+          throw error;
```
moodmosaic, line 299: Don't we want to `radio.emit` the error message? Also, if this fails, do we want to keep going?

property.ts
```diff
@@ -138,7 +139,16 @@ export const checkProperties = (
               functionArgsArb: fc.tuple(...functionArgsArb),
             })
             .map((args) => ({ ...r, ...args }));
-        }),
+        })
+        .chain((r) =>
+          fc
+            .record({
+              burnBlocksToMine: r.mineBurnBlocks
```
moodmosaic, line 146: Same comment on naming as above.

property.ts
```diff
@@ -174,11 +184,13 @@ export const checkProperties = (
         if (discarded) {
           radio.emit(
             "logMessage",
-            `${dim(testCallerWallet)} ${yellow("[WARN]")} ${
-              r.testContractId.split(".")[1]
-            } ${underline(r.selectedTestFunction.name)} ${dim(
-              printedTestFunctionArgs
-            )}`
+            `₿ ${simnet.burnBlockHeight.toString().padStart(6)} ` +
```
moodmosaic, line 187: Same comment on padding as above.

invariant.ts
BowTiedRadone: Yes. Otherwise, after the Stacks and Bitcoin height addition the output would look horrible.

invariant.ts
BowTiedRadone: Yes. `canMineBlocks` will work.

invariant.ts
BowTiedRadone: In 100 runs, the Bitcoin block gets up to a roughly average of 270.

invariant.ts
moodmosaic: 🫡

invariant.ts
BowTiedRadone: Perhaps 8 will be the right choice for 1M runs. `1.000.000 / 100 * 270 = 2.700.000` (where 270 is average blocks mined in 100 runs). Considering the fact we can go up to 2M runs, 8 should be ok.

BowTiedRadone: Good catch!

invariant.ts
moodmosaic: I was thinking to define the `max` based on https://fast-check.dev/docs/configuration/larger-entries-by-default/#override-the-default-size.

invariant.ts
moodmosaic: OK, then if that fails, we basically stop (`heatstroke`). Why don't we even need to wrap in a `try-catch` then?

invariant.ts
BowTiedRadone: Wouldn't we want to achieve an `inverse proportion` here? Mine more blocks in the beginning to cover more burnchain-dependent state transitions for small `numRuns`, and as `numRuns` increases mine less? IMO having a finite `max` fits our scenario, but I am open to trying the suggestion you brought up.

BowTiedRadone: For example, to achieve an inverse proportion, max can be `Math.floor(10M / numRuns)`, and still be deterministic.

BowTiedRadone: In fact, `Math.ceil` would be better to avoid errors in case min is 1 and `numRuns > 10M`.

moodmosaic: I’d be curious to see how that arbitrary looks like when we sample it. Perhaps you’re right, but let’s verify it first by sampling?

invariant.ts
moodmosaic: Once we have a static part on the TUI we won’t have to worry about passing.

invariant.ts
BowTiedRadone: Using my proposed formula:

When `numRuns=100`, some 10-items samples looks like this:
```bash
#1

[592,17730,2,41599,7,16,99999,10,99997,14]

#2

[77346,100000,63091,9,11,99995,17433,22496,1,99999]

#3

[44555,55644,8,12,10,14,24745,1,99989,99989]
```

For a medium `numRuns` (`500k` in this case):

```bash
# 1

[18,11,5,2,19,5,9,7,2,20]

#2

[18,20,14,12,13,2,11,11,14,16]
```

For a high `numRuns` (`2638912739812` in this case, which is greater than the 10M multiplier):

```bash
# Always a list of 1s

[1,1,1,1,1,1,1,1,1,1]
```

invariant.ts
BowTiedRadone: The updated inverse proportion approach seems to fit the 6 chars limit and there is also some buffer.

property.ts
BowTiedRadone: See https://github.com/stacks-network/rendezvous/pull/81#discussion_r1905351389.

invariant.ts
BowTiedRadone: The padding was updated to 8 chars in e846c3f, increasing the buffer for long runs.

moodmosaic (approved):
🚀

Issue #65, opened 2024-12-17
Support for traits


moodmosaic:
See https://github.com/stacks-network/rendezvous/issues/7#issuecomment-2544137855.

moodmosaic:
Note https://github.com/stacks-network/rendezvous/pull/83#discussion_r1907275506 when implementing this.

PR #83, opened 2025-01-08
Skip functions with `trait reference` parameters from selection

BowTiedRadone:
This PR updates `rv` by gracefully handling `trait references`. It skips functions with `trait reference` parameters from being selected for argument generation and random execution, which previously caused crashes. Currently, `trait references` are not supported, but this feature is planned (see #65).

Additionally, this PR follows [The Boy Scout Rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html), leaving the codebase cleaner than it was found. _Apparently_ unrelated updates such as refined arbitrary definitions, refactored comments, and updated logging can be found in this PR.

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -4,6 +4,7 @@ import { EventEmitter } from "events";
 import { checkProperties } from "./property";
 import { checkInvariants } from "./invariant";
 import {
+  getContractNameFromContractId,
   getFunctionsFromContractInterfaces,
   getSimnetDeployerContractsInterfaces,
 } from "./shared";
@@ -117,7 +118,7 @@ export async function main() {
   const rendezvousList = Array.from(
     getSimnetDeployerContractsInterfaces(simnet).keys()
   ).filter((deployedContract) =>
-    [sutContractName].includes(deployedContract.split(".")[1])
+    [sutContractName].includes(getContractNameFromContractId(deployedContract))
   );
 
   const rendezvousAllFunctions = getFunctionsFromContractInterfaces(
--- a/app.ts
+++ b/app.ts
@@ -9,6 +9,7 @@ import {
 } from "./shared";
 import { issueFirstClassCitizenship } from "./citizen";
 import { version } from "./package.json";
+import { red } from "ansicolor";
 
 const logger = (log: string, logLevel: "log" | "error" | "info" = "log") => {
   console[logLevel](log);
@@ -42,7 +43,7 @@ const parseOptionalArgument = (argName: string) => {
 export async function main() {
   const radio = new EventEmitter();
   radio.on("logMessage", (log) => logger(log));
-  radio.on("logFailure", (log) => logger(log, "error"));
+  radio.on("logFailure", (log) => logger(red(log), "error"));
 
   const args = process.argv;
   if (args.includes("--help")) {
@@ -55,7 +56,9 @@ export async function main() {
   if (!manifestDir || manifestDir.startsWith("--")) {
     radio.emit(
       "logMessage",
-      "\nNo path to Clarinet project provided. Supply it immediately or face the relentless scrutiny of your contract's vulnerabilities."
+      red(
+        "\nNo path to Clarinet project provided. Supply it immediately or face the relentless scrutiny of your contract's vulnerabilities."
+      )
     );
     radio.emit("logMessage", helpMessage);
     return;
@@ -66,7 +69,9 @@ export async function main() {
   if (!sutContractName || sutContractName.startsWith("--")) {
     radio.emit(
       "logMessage",
-      "\nNo target contract name provided. Please provide the contract name to be fuzzed."
+      red(
+        "\nNo target contract name provided. Please provide the contract name to be fuzzed."
+      )
     );
     radio.emit("logMessage", helpMessage);
     return;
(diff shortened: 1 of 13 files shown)


shared.ts
```diff
@@ -303,3 +307,38 @@ const complexTypesToCV: ComplexTypesToCV = {
...
+ */
+export const isTraitReferenceFunction = (
+  fn: ContractInterfaceFunction
+): boolean => {
+  const hasTraitReference = (type: ParameterType): boolean => {
+    if (typeof type === "string") {
+      // The type is a base type.
+      return type === "trait_reference";
+    } else {
+      // The type is a complex type.
```
moodmosaic, line 325: Are all cases in this `else` branch covered by tests? To avoid surprises in the wild.

invariant.ts
```diff
@@ -53,6 +54,68 @@ export const checkInvariants = (
...
+    return;
+  }
+
+  const eligibleFunctions = functions.filter(
+    (fn) => !isTraitReferenceFunction(fn)
+  );
+
+  const eligibleInvariants = invariantFunctions.filter(
+    (fn) => !isTraitReferenceFunction(fn)
+  );
```
moodmosaic, line 97: If I understand correctly, when we add support for traits, this filtering will be omitted. One thing to consider, though: if we end up supporting traits via #65, we’ll need a way to dynamically omit it at runtime. Not a blocker for merging, just something to keep in mind.

moodmosaic (commented):
Looking good! Makes the overall arbitrary setup cleaner also! Had some comments on the
- removal of getContractNameFromRendezvousId
- coverage of isTraitReferenceFunction

invariant.ts
```diff
@@ -61,69 +124,36 @@ export const checkInvariants = (
...
-
-          return fc
+        .chain((r) =>
+          fc
             .record({
-              selectedFunction: functionArbitrary,
-              selectedInvariant: invariantFunctionArbitrary,
+              selectedFunction: fc.constantFrom(...eligibleFunctions),
+              // FIXME: For invariants, we have to be able to select a random
+              // number of them (zero or more).
```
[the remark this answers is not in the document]
BowTiedRadone: The `FIXME` is an old one. What I did was to preserve it.

invariant.ts
```diff
@@ -354,14 +378,6 @@ export const initializeClarityContext = (
     });
   });
 
-/**
- * Get the contract name from the Rendezvous identifier.
- * @param rendezvousId The Rendezvous contract identifier.
- * @returns The contract name.
- */
-export const getContractNameFromRendezvousId = (rendezvousId: string) =>
```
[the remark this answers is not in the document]
BowTiedRadone: Exactly. This method was no longer needed, since we work with the original contract and not with `_rendezvous` one. It is now inlined because it is a simple split, and the `replace` was not doing anything.

invariant.ts
BowTiedRadone: Why would we want to omit at runtime? In case the user does not specify what traits to use in the configuration?

invariant.ts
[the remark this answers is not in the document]
BowTiedRadone: I can remove it now.

shared.ts
BowTiedRadone: No. I will figure out a test to cover all the possible types!

invariant.ts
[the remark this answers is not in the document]
BowTiedRadone: A new method was now added. It is called `getContractNameFromContractId` and does what you highlighted. And yes, you were right. The rule of three kicked in.

shared.ts
BowTiedRadone: Done in d315e66384329599de8534d22b3a21c4707eb479. ✅

invariant.ts
moodmosaic: Because here we’re telling the core, “hey, if you find a trait type then skip it, regardless of whether you can generate an instance of that type”.

BowTiedRadone: Oh, sure! This part of the infrastructure **is temporary and will be removed once trait references are supported**. After `trait references` are supported, we may want to identify the trait reference the function requires, and see if there is any contract in the `Clarinet` project implementing that trait. If not, skip it and report to the user. This could happen as a `pre-check`, or just `post-execution`, mentioning the skipped `functions/invariants/tests` along with the reason.

moodmosaic: Agreed. Good that we’re < version 1 for this kind of stuff 🚀🛸

app.ts
```diff
@@ -110,7 +118,7 @@ export async function main() {
   const rendezvousList = Array.from(
     getSimnetDeployerContractsInterfaces(simnet).keys()
   ).filter((deployedContract) =>
-    [sutContractName].includes(deployedContract.split(".")[1])
+    [sutContractName].includes(getContractNameFromContractId(deployedContract))
```
moodmosaic, line 121: That's cleaner 👍

invariant.ts
```diff
@@ -61,69 +124,34 @@ export const checkInvariants = (
     fc.property(
       fc
         .record({
-          rendezvousContractId: fc.constantFrom(...rendezvousList),
+          // The target contract identifier. It is a constant value equal
+          // to the first contract in the list. The arbitrary is still needed,
+          // being used for reporting purposes in `heatstroke.ts`.
```
moodmosaic, line 129: Up to you, I think this comment is no longer needed.

shared.tests.ts
```diff
@@ -74,6 +77,331 @@ describe("Simnet contracts operations", () => {
...
+      outputs: { type: "bool" },
+    } as ContractInterfaceFunction;
+
+    // Act
+    const result = isTraitReferenceFunction(unrecognizedTypeFunction);
+
+    // Assert
+    expect(result).toBe(false);
+  });
+});
```
moodmosaic, line 403: Nice test-suite there👍 💯

moodmosaic (approved):
🚀 💯

PR #88, opened 2025-01-13
Select contracts implementing the correct trait from the Clarinet project for argument generation

### LINKED ISSUE #65: Support for traits

BowTiedRadone:
This PR adds support for arbitrary selection of `trait reference` parameters. The new logic for handling trait implementations in trait reference parameters is as follows: 

- If a trait reference parameter is found in public functions or invariants during invariant testing, or in test functions during property testing, Rendezvous will scan the user's Clarinet project for a contract that implements the referenced trait.
- If trait implementations are found for that specific trait reference, they will be added to a selection pool. Rendezvous uses fast-check to randomly select one of them for each run, ensuring seed-based randomness.

To test contracts with trait reference parameters, users must select at least one trait implementation for the referenced trait and add it to their Clarinet project, either as a direct contract or a project requirement.

This PR resolves #65.


property.ts
```diff
@@ -23,11 +32,195 @@ export const checkProperties = (
...
+        return f;
+      }),
+    ])
+  );
+
+  console.log(
+    "Enriched rendezvousAllFunctions:",
+    JSON.stringify(enrichedRendezvousAllFunctions.get(testContractId), null, 2)
+  );
+
```
moodmosaic, line 218: As a first step I would extract this into a test.

property.ts
```diff
@@ -23,11 +32,195 @@ export const checkProperties = (
...
+  //           ok: "bool",
+  //           error: "none",
+  //         },
+  //       },
+  //     },
+  //   },
+  // ];
+
+  // Make a copy map with the enriched rendezvousAllFunctions object.
+  const enrichedRendezvousAllFunctions = new Map(
```
moodmosaic, line 186: What does it mean?

shared.types.ts
```diff
@@ -17,6 +22,24 @@ import fc from "fast-check";
 
 // Types used for Clarity Value conversion.
 
+export type EnrichedContractInterfaceFunction = {
```
moodmosaic, line 25: I see now. Why do we need this union?

shared.types.ts
```diff
@@ -59,7 +78,8 @@ type ComplexType =
   | { list: { type: ParameterType; length: number } }
   | { tuple: { name: string; type: ParameterType }[] }
   | { optional: ParameterType }
-  | { response: { ok: ParameterType; error: ParameterType } };
+  | { response: { ok: ParameterType; error: ParameterType } }
+  | { trait_reference: { name: string } };
```
moodmosaic, line 82: Is just `name` enough (e.g. if we have multiple traits).

moodmosaic (commented):
Interesting work on the draft implementation so far! Left some comments.  

The next steps I see are:  

- Create (perhaps parameterized?) test(s) based on the draft/example in `property.ts`.  
- Add as many more test cases as possible to expand coverage.
- Tweak the draft implementation to ensure all tests pass.  
- (Once tests provide decent coverage, we'll consider refactoring with a  
  *parsec*-style library.)

ast1737030528990.json
```diff
@@ -0,0 +1,963 @@
...
+                          "end_column": 33
+                        }
+                      },
+                      {
+                        "expr": {
+                          "TraitReference": [
+                            "ft-trait",
+                            {
+                              "Imported": {
+                                "name": "sip-010-trait",
```
moodmosaic, line 844: This is the trait type, in a way, right?

example/contracts/trait.clar
```diff
@@ -0,0 +1,5 @@
+(use-trait ft-trait 'SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.sip-010-trait-ft-standard.sip-010-trait)
+
+(define-public (trait-transfer-function (token <ft-trait>))
```
moodmosaic, line 3: Here do we always pick the trait impl from the file (e.g. the one on line 1)?

property.ts
```diff
@@ -11,7 +11,16 @@ import {
...
+import {
+  ContractInterfaceFunction,
+  ContractInterfaceFunctionAccess,
+  ContractInterfaceFunctionArg,
+  ContractInterfaceFunctionOutput,
+  List,
+  TraitReference,
+} from "@hirosystems/clarinet-sdk-wasm";
+import { writeFileSync } from "fs";
+import { EnrichedContractInterfaceFunction } from "./shared.types";
```
moodmosaic, line 23: I keep not being able to fully follow the purpose of `EnrichedContractInterfaceFunction`. Is it a union?

property.ts
```diff
@@ -23,13 +32,199 @@ export const checkProperties = (
...
+  //           error: "none",
+  //         },
+  //       },
+  //     },
+  //   },
+  // ];
+
+  // Make a copy map with the enriched rendezvousAllFunctions object.
+  const enrichedRendezvousAllFunctions = new Map(
+    Array.from(rendezvousAllFunctions, ([contractId, functions]) => [
```
moodmosaic, line 187: I see what's going on here. Basically the `rendezvousAllFunctions` list (correctly) misses the trait ref/impl (in your case `traitReferenceName` and `traitReferenceImport`).

What is the advantage of doing a union? AFAICT, `enrichedRendezvousAllFunctions` is that union.

moodmosaic: From a high level perspective, this is the structure we're dealing with, right?

```ts
type TraitReference = {
  name: string;
  import: string;
};

type Argument = {
  name: string;
  type: {
    trait_reference?: TraitReference;
  };
};

type FunctionInfo = {
  name: string;
  args: Argument[];
};

type ContractFunctions = [string, FunctionInfo[]];
```

moodmosaic (commented):
Left questions and comments at the abstraction level. Let's make sure we get those right before proceeding.

property.ts
BowTiedRadone: When retrieving the contract interface using Simnet, if a trait is found, the type is only displayed as a generic string "`trait_reference`". For example, here’s the interface for the [trait-transfer-function](https://github.com/stacks-network/rendezvous/blob/74a086b69200422ccf086bebb292614feb353ee4/example/contracts/trait.clar#L3):

```
{
  name: "trait-transfer-function",
  access: "public",
  args: [{ name: "token", type: "trait_reference" }],
  outputs: { type: { response: [Object] } }
}
```
Our utilities, which handle argument generation, use the function interface to build arbitraries recursively. However, if the interface is passed as-is, there isn’t enough context to determine which specific trait the `trait_reference` refers to.

To address this, when a `trait_reference` is found in the function interface, we search for that parameter (at any nesting depth) in the contract’s AST. We then enrich the function interface with the detailed trait information and pass the updated interface to the arbitraries utility.

During argument generation, the enriched trait parameter is used. Ideally, we will have a pre-populated record or map of all contracts implementing various traits. If we find contracts implementing the required trait, one is selected at random. When converting this into a `ClarityValue`, the selected contract identifier is simply turned into a `principalCV`.

ast1737030528990.json
BowTiedRadone: Yes. This is the **_imported_** trait type. The type for an **_implemented_** trait is slightly different. The difference between the both can be observed [here](https://github.com/stacks-network/rendezvous/blob/74a086b69200422ccf086bebb292614feb353ee4/shared.ts#L441-L445), where they are compared.

example/contracts/trait.clar
BowTiedRadone: Yes, with the mention that there can be multiple traits. Do you think we could search directly in the file instead of parsing the AST? Now that you highlighted it, I realise it could be an option.

BowTiedRadone: There's only one thing to be considered: we would still have to parse the AST to find contracts that implement that specific trait. That would also require turning the principal in a bytes array for comparison.

moodmosaic: It'd be more robust to stick to the AST as the syntax in the file might vary.

>with the mention that there can be multiple traits.

How can it possibly be? Is it at the function level or contract level?

property.ts
moodmosaic: Nice 👍

invariant.ts
```diff
@@ -144,10 +144,18 @@ export const checkInvariants = (
           fc
             .record({
               functionArgsArb: fc.tuple(
-                ...functionToArbitrary(r.selectedFunction, simnetAddresses)
+                ...functionToArbitrary(
+                  r.selectedFunction,
+                  simnetAddresses,
+                  simnet
```
moodmosaic, line 150: Do we need multiple stuff from the simnet object? Or just one?

invariant.ts
```diff
@@ -144,10 +144,18 @@ export const checkInvariants = (
...
+                )
               ),
               invariantArgsArb: fc.tuple(
-                ...functionToArbitrary(r.selectedInvariant, simnetAddresses)
+                ...functionToArbitrary(
+                  r.selectedInvariant,
+                  simnetAddresses,
+                  simnet
```
moodmosaic, line 157: Do we need multiple stuff from the simnet object? Or just one?

property.ts
```diff
@@ -8,10 +8,13 @@ import {
   functionToArbitrary,
   getContractNameFromContractId,
   getFunctionsListForContract,
+  getTraitReferenceData,
   isTraitReferenceFunction,
 } from "./shared";
 import { dim, green, red, underline, yellow } from "ansicolor";
 import { ContractInterfaceFunction } from "@hirosystems/clarinet-sdk-wasm";
+import { writeFileSync } from "fs";
```
moodmosaic: I guess this is redundant import?

property.ts
```diff
@@ -23,13 +26,233 @@ export const checkProperties = (
   runs: number | undefined,
   radio: EventEmitter
 ) => {
+  const testContractId = rendezvousList[0];
+
+  // console.log(rendezvousAllFunctions.get(testContractId)![2]);
```
moodmosaic: Time to get rid of those comments?

property.ts
```diff
@@ -23,13 +26,233 @@ export const checkProperties = (
...
+  //         "response": {
+  //           "ok": "bool",
+  //           "error": "none"
+  //         }
+  //       }
+  //     }
+  //   },
+  // ]
+  const functionInterface = rendezvousAllFunctions.get(testContractId)!;
+  // writeFileSync(
```
moodmosaic: Do we need this commented-out?

property.ts
```diff
@@ -23,13 +26,233 @@ export const checkProperties = (
...
+  //   "./fixtures/direct-trait-interface.json",
+  //   JSON.stringify(functionInterface, null, 2)
+  // );
+
+  // We know that the "token" parameter of the "test-trait" function is a trait
+  // reference. We need to find out which trait is being referenced. For this,
+  // we need to analyze the AST and find the trait reference.
+  const ast = simnet.getContractAST(testContractId.split(".")[1]);
+
+  // writeFileSync(
```
moodmosaic: Same here.

fixtures/direct-trait-2nd-parameter-ast.json
```diff
@@ -0,0 +1,928 @@
+{
```
moodmosaic, line 1: What does a *fixture* means in our context?

traits.ts
```diff
@@ -0,0 +1,302 @@
...
+              },
+            },
+          };
+        }
+      }
+      return arg;
+    });
+  };
+
+  // Iterate over all functions in the target contract
```
moodmosaic, line 63:
```suggestion
  // Iterate over all functions in the target contract.
```

traits.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  enriched.set(targetContractId, enrichedFunctions);
+  return enriched;
+};
+
+export const getTraitReferenceData = (
+  ast: IContractAST,
+  functionName: string,
+  paramPath: string[]
+): [string, any] | [undefined, undefined] => {
```
moodmosaic, line 79: Can we document this return type?

traits.ts
```diff
@@ -0,0 +1,302 @@
...
+  const findTraitReference = (
+    paramList: any[],
```
moodmosaic, line 81:
```suggestion
    args: any[],
```
?

traits.ts
```diff
@@ -0,0 +1,302 @@
...
+    path: string[]
+  ): [string, any] | [undefined, undefined] => {
```
moodmosaic, line 83: Same here, can we document this return type? It looks complex.

traits.ts
```diff
@@ -0,0 +1,302 @@
...
+    for (const param of paramList) {
+      if (!param.expr || !(param.expr as List).List) continue;
```
moodmosaic, line 85:
```suggestion
      if (!param.expr || !(param.expr as List).List) { continue; }
```

Would you prefer to stick to curly-braces everywhere? It's easy to mis-indent code and have unexpected side-effects:

```ts
if (condition)
    doSomething();
    doAnotherThing(); // Always executes, regardless of the condition.
```

vs

```ts
if (condition) {
    doSomething();
    doAnotherThing();
}
```

traits.ts
```diff
@@ -0,0 +1,302 @@
...
+            traitReferenceNode &&
+            (traitReferenceNode.expr as TraitReference).TraitReference
+          ) {
+            const [name, importData] = (
+              traitReferenceNode.expr as TraitReference
+            ).TraitReference;
+            return [name, importData];
+          }
+        } else {
+          // Recursively check for nested trait references
```
moodmosaic, line 104:
```suggestion
          // Recursively check for nested trait references.
```

traits.ts
```diff
@@ -0,0 +1,302 @@
...
+): [string, any] | [undefined, undefined] => {
+  const findTraitReference = (
+    paramList: any[],
+    path: string[]
+  ): [string, any] | [undefined, undefined] => {
+    for (const param of paramList) {
+      if (!param.expr || !(param.expr as List).List) continue;
+
+      const paramNameNode = (param.expr as List).List[0];
+      if (!paramNameNode || !(paramNameNode.expr as Atom).Atom) continue;
```
moodmosaic, line 88:
```suggestion
      if (!paramNameNode || !(paramNameNode.expr as Atom).Atom) { continue; }
```

moodmosaic: (Same for all occurrences.)

traits.types.ts
```diff
@@ -0,0 +1 @@
+// This file is a placeholder for the traits processing-related types.
```
moodmosaic, line 1: Is this a placeholder for those functions returning `[string, any] | [undefined, undefined]`?

traits.tests.ts
```diff
@@ -0,0 +1,873 @@
...
+  enrichInterfaceWithTraitData,
+  getContractIdsImplementingTrait,
+  isTraitReferenceFunction,
+} from "./traits";
+
+describe("Trait reference processing", () => {
+  it("correctly builds the trait reference map for a direct trait that is the first parameter", () => {
+    // Arrange
+    const allFunctionsInterfaces = JSON.parse(
+      readFileSync("./fixtures/direct-trait-functions-interfaces.json", "utf-8")
```
moodmosaic, line 16: What is a *fixture* really? I used to contribute to [*autofixture*](https://github.com/AutoFixture/AutoFixture/graphs/contributors), is it going to end up being something related?

traits.tests.ts
```diff
@@ -0,0 +1,873 @@
...
+    );
+
+    const expectedTraitReferenceMap = new Map(
```
moodmosaic, line 19:
```suggestion
    const expected = new Map(
```

traits.tests.ts
```diff
@@ -0,0 +1,873 @@
...
+      Object.entries({
+        "test-trait": { token: "trait_reference" },
+      })
+    );
+
+    // Act
+    const traitReferenceMap = buildTraitReferenceMap(allFunctionsInterfaces);
```
moodmosaic, line 26:
```suggestion
    const actual = buildTraitReferenceMap(allFunctionsInterfaces);
```

moodmosaic: How about using some generic naming here? It'll make the rest of the tests easier to follow when reading them after.

traits.tests.ts
```diff
@@ -0,0 +1,873 @@
...
+
+    // Assert
+    expect(traitReferenceMap).toEqual(expectedTraitReferenceMap);
+  });
+
+  it("correctly builds the trait reference map for a direct trait that is the second parameter", () => {
+    // Arrange
+    const allFunctionsInterfaces = JSON.parse(
+      readFileSync(
+        "./fixtures/direct-trait-2nd-parameter-functions-interfaces.json",
```
moodmosaic, line 36: Instead of reading those from a file, would it be possible to inline the json (or the part of the json we're interested at) here in the test's arrange phase?

traits.tests.ts
```diff
@@ -0,0 +1,873 @@
...
+
+    // Assert
+    expect(traitReferenceMap).toEqual(expectedTraitReferenceMap);
+  });
+
+  it("correctly builds the trait reference map for a direct trait that is the fifth parameter", () => {
+    // Arrange
+    const allFunctionsInterfaces = JSON.parse(
+      readFileSync(
+        "./fixtures/direct-trait-5th-parameter-functions-interfaces.json",
```
moodmosaic, line 58: Could we fabricate the json so it dynamically moves the trait parameter? This way we could easily test of to seven<sup>1</sup>.

---

<sup>1</sup>*Miller, G. A. (1956). "The Magical Number Seven, Plus or Minus Two: Some Limits on Our Capacity for Processing Information." Psychological Review, 63(2), 81–97.*

traits.tests.ts
```diff
@@ -0,0 +1,873 @@
...
+    const traitReferenceMap = buildTraitReferenceMap(allFunctionsInterfaces);
+
+    const ast = JSON.parse(
+      readFileSync("./fixtures/direct-trait-ast.json", "utf-8")
+    );
+
+    const expected = new Map(
+      Object.entries({
+        [targetContractId]: [
+          {
```
moodmosaic, line 117: For example, here it's very nice that the reader can see the json without searching in files.

invariant.ts
BowTiedRadone: Currently, matching traits are selected during argument generation. This approach is WIP and inefficient, but it works for now. Ideally, once all trait reference nesting types are handled, we’ll pre-populate a map that links stringified trait data to the contracts implementing it. This will eliminate the need to pass simnet, but the pre-populated map.

ast1737030528990.json
BowTiedRadone: The difference is as follows: 

Contracts that **implement** one or more traits will have the trait data specified in the `implemented_traits` field at the top level of the AST. This is how we identify contracts that implement a specific trait. 

Contracts that **use** the trait and may have functions with `trait_reference` parameters will have the trait data nested at the parameter level within the AST. These two types of trait data have different structures:

**Implemented:**
```json
"implemented_traits": [
  {
    "name": "sip-010-trait",
    "contract_identifier": {
      "issuer": [
        22,
        [
          9, 159, 184, 137, 38, 216, 47, 48, 178,
          244, 14, 175, 62, 228, 35, 203, 114, 91,
          219, 59
        ]
      ],
      "name": "sip-010-trait-ft-standard"
    }
  }
]
```

**Imported (nested at the parameter level):**
```json
"TraitReference": [
  "ft-trait",
  {
    "Imported": {
      "name": "sip-010-trait",
      "contract_identifier": {
        "issuer": [
          22,
          [
            9, 159, 184, 137, 38, 216, 47, 48, 178,
            244, 14, 175, 62, 228, 35, 203, 114, 91,
            219, 59
          ]
        ],
        "name": "sip-010-trait-ft-standard"
      }
    }
  }
]
```

example/contracts/trait.clar
BowTiedRadone: I’ve re-read your first comment and realized I misunderstood it—sorry about that! Here’s how it works now:

When selecting a function, we first check if it has `trait_reference` parameters at any depth (direct and/or nested within other complex types). These are easy to identify in the function interface because their type is mapped to `trait_reference`. 

If such parameters are found, we create a map linking the function name to the paths where the trait references appear. These paths allow us to traverse the AST and retrieve all the trait imports data.

The retrieved trait import data replaces the plain `trait_reference` type in the function interface. This enriched data is then used during argument generation to identify and select all contracts that implement the same `trait_reference` as the parameter being generated.

For a clearer understanding of how the paths map looks like, you can check the tests [here](https://github.com/stacks-network/rendezvous/blob/3e8743c774229576c60975bf05b93f8e66a9a264/traits.tests.ts#L13-L122).

invariant.ts
BowTiedRadone: Currently we need to get any project contract's AST.

traits.ts
BowTiedRadone: Sure!

traits.types.ts
BowTiedRadone: Exactly!

traits.tests.ts
BowTiedRadone: Having access to the full AST provides a better understanding of its structure and allows us to design and understand the trait-related utility functions more effectively. Since the AST for a contract with just four simple functions is already 882 lines long, I would definitely not inline the ASTs in the tests.

invariant.ts
moodmosaic: Couldn't we pass just that function then from the simnet? E.g. pass `simnet.getContractAST` as a delegate.

invariant.ts
moodmosaic: Couldn't we pass just that function then from the simnet? E.g. pass `simnet.getContractAST` as a delegate.

property.ts
```diff
@@ -94,10 +129,8 @@ export const checkProperties = (
 
   const simnetAddresses = Array.from(simnetAccounts.values());
 
-  const testContractId = rendezvousList[0];
-
   const testFunctions = getFunctionsListForContract(
-    testContractsTestFunctions,
+    enrichedTestFunctionsInterfacesMap,
```
moodmosaic, line 133: Adding this as a note for us to consider alternatives, otherwise "enriched" it is.

property.ts
```diff
@@ -151,7 +168,11 @@ export const checkProperties = (
           fc
             .record({
               functionArgsArb: fc.tuple(
-                ...functionToArbitrary(r.selectedTestFunction, simnetAddresses)
+                ...functionToArbitrary(
+                  r.selectedTestFunction,
+                  simnetAddresses,
+                  simnet
```
moodmosaic, line 174: Same comment about passing something more specific from the simnet here instead of the whole instance.

shared.ts
```diff
@@ -1,12 +1,14 @@
 import fc from "fast-check";
 import {
-  BaseType,
+  BaseTypeAfterEnrich,
```
moodmosaic, line 3: Could be at least EnrichedType? or EnrichedBaseType? In case we go with "enriched".

shared.ts
```diff
@@ -54,20 +58,25 @@ export const getFunctionsFromContractInterfaces = (
   );
 
 export const getFunctionsListForContract = (
-  functionsMap: Map<string, ContractInterfaceFunction[]>,
+  functionsMap: Map<string, any[]>,
```
moodmosaic, line 61: Shouldn't `any` be `EnrichedXyz`?

traits.ts
```diff
@@ -0,0 +1,388 @@
...
+              },
+            },
+          };
+        }
+      }
+      return arg;
+    });
+  };
+
+  // Iterate over all functions in the target contract
```
moodmosaic, line 116:
```suggestion
  // Iterate over all functions in the target contract.
```

traits.ts
moodmosaic: Is it documented?

moodmosaic (commented):
This is massive 🥇 After refactoring `simnet` and coming up with candidate alternatives for "enrich" this should be ready for review.

traits.tests.ts
BowTiedRadone: Resolved.

example/Clarinet.toml
```diff
@@ -3,6 +3,9 @@ name = "example"
 telemetry = false
 cache_dir = "./.cache"
 
+[[project.requirements]]
+contract_id = 'SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.ststx-token'
```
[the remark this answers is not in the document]
BowTiedRadone: That's right. Before merging, we should decide:
- what contract that imports traits we want to add to the example. Can be even handmade, but should highlight the functionality
- what contract that implements traits we want to use. Same here, can be either handmade and added to the project as a new contract, or added as a requirement.

example/contracts/trait.tests.clar
```diff
@@ -0,0 +1,9 @@
+(use-trait ft-trait 'SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.sip-010-trait-ft-standard.sip-010-trait)
```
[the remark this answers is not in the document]
BowTiedRadone: What happens is that when [this function](https://github.com/stacks-network/rendezvous/blob/47a425b1a157a2aa0591328a7270115a57bbc1eb/example/contracts/trait.tests.clar#L3) is picked, `ststx-token` is the contract that can be picked as an argument, because it is the only one in the project (requirements included) that implements the trait.

shared.tests.ts
```diff
@@ -77,331 +75,6 @@ describe("Simnet contracts operations", () => {
   });
 });
 
-describe("Trait reference detection", () => {
```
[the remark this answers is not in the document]
BowTiedRadone: Yes, they are.

shared.ts
```diff
@@ -172,38 +191,56 @@ const complexTypesToArbitrary: ComplexTypesToArbitrary = {
       maxLength: length,
     }),
   "string-utf8": (length: number) => fc.string({ maxLength: length }),
-  list: (type: ParameterType, length: number, addresses: string[]) =>
-    fc.array(parameterTypeToArbitrary(type, addresses), { maxLength: length }),
+  list: (
+    type: EnrichedParameterType,
+    length: number,
+    addresses: string[],
+    simnet: Simnet
```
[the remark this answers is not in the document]
BowTiedRadone: Yes, the map will be used also here.

shared.types.ts
```diff
@@ -41,27 +62,45 @@ export type ComplexTypesToCV = {
...
-  | "uint128"
+export type BaseTypeBeforeEnrich =
+  | "int"
+  | "uint"
   | "bool"
   | "principal"
   | "trait_reference";
 
-type ComplexType =
+export type BaseTypeAfterEnrich = "int128" | "uint128" | "bool" | "principal";
```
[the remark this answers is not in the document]
BowTiedRadone: No, because BaseType initially has also "`trait_reference`".

shared.ts
```diff
@@ -172,38 +191,56 @@ const complexTypesToArbitrary: ComplexTypesToArbitrary = {
...
         status: fc.constant("error"),
-        value: parameterTypeToArbitrary(errType, addresses),
+        value: parameterTypeToArbitrary(errType, addresses, simnet),
       })
     ),
+  trait_reference: (traitData: TraitImportType, simnet: Simnet) => {
+    return fc.constantFrom(
+      ...getContractIdsImplementingTrait(traitData, simnet)
+    );
+  },
```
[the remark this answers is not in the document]
BowTiedRadone: Resolved ✅

BowTiedRadone:
Gigantic cleaning session happened here. At least a couple things to decide on before merging:
- Which contract that imports traits should be added to the example folder? It could be a handmade contract, but it should showcase the functionality clearly.
- Which contract that implements traits do we want to use? This can either be a handmade contract added to the project or specified as a requirement.

Ideally, we should have two trait implementations for the same trait, so Rendezvous can randomly pick one of them for any trait reference parameter.

example/Clarinet.toml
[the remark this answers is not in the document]
BowTiedRadone: [This](https://github.com/stacks-network/rendezvous/pull/88#issuecomment-2611175099) being said, I will resolve this comment. https://github.com/stacks-network/rendezvous/pull/88#discussion_r1924418867 remains unresolved for the moment, as it points to the same discussion topic.

moodmosaic:
>Which contract that imports traits should be added to the example folder? It could be a handmade contract, but it should showcase the functionality clearly.

Any contract that shows a real-world use case and stays under 100 lines.

>Which contract that implements traits do we want to use? This can either be a handmade contract added to the project or specified as a requirement.
 
I'd say both: a local handmade one and one added as a requirement.

moodmosaic (approved):
This is a massive and tricky-to-implement PR. It's been carefully crafted, and I want to leave an approval noting that it's one of the most well-crafted PRs (of this size) I've reviewed in over a decade.

I've left a few comments, but they aren't blockers. It would be nice to address them, but we could merge and handle those separately if needed.

One thing to consider for this PR is whether to keep the full commit history or squash it into a single commit. Keeping the history preserves individual commit context, which can be helpful for debugging and provides a detailed timeline of changes. On the other hand, it can make the history noisier with too many intermediate commits.

Squashing into a single commit results in a cleaner, more concise history, making it easier to track high-level changes. It also simplifies `git revert` if we need to roll back due to blockers, which is especially useful since we're still < v1 and can afford breaking changes. However, squashing does lose the detailed commit trail, which can sometimes be useful for debugging.

As Linus Torvalds put it: "I want clean history, but that really means (a) clean and (b) history." (Source: https://www.mail-archive.com/dri-devel@lists.sourceforge.net/msg39091.html)

citizen.ts
```diff
@@ -111,7 +111,7 @@ export const groupContractsByEpochFromSimnetPlan = (
 
 /**
  * Deploy the contracts to the simnet in the correct order.
- * @param simnet - The simnet instance.
+ * @param simnet The simnet instance.
```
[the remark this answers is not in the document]
BowTiedRadone: Checked all "`The simnet instance`", and removed the dash for consistency between all the other occurences.

example/.cache/requirements/SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.sip-010-trait-ft-standard.clar
```diff
@@ -0,0 +1,24 @@
+(define-trait sip-010-trait
+  (
+    ;; Transfer from the caller to a new principal
+    (transfer (uint principal principal (optional (buff 34))) (response bool uint))
+
+    ;; the human readable name of the token
```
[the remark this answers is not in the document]
BowTiedRadone: Can we update the `.cache`? Isn't it overwritten every run?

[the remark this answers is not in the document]
moodmosaic: Can we just .gitignore it?

invariant.ts
```diff
@@ -35,11 +41,73 @@ export const checkInvariants = (
     rendezvousAllFunctions
   );
 
+  // The Rendezvous identifier is the first one in the list. Only one contract
+  // can be fuzzed at a time.
```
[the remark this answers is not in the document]
BowTiedRadone: `rendezvousList` can also become a string representing the target contract. It is left like that for the case we'll ever want to re-introduce support for fuzzing multiple contracts at once.

traits.types.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * The trait reference structure, as it appears in the AST under the
+ * `implemented_traits` field.
+ */
+export type ImplementedTraitType = {
+  name: string;
+  contract_identifier: { issuer: number[]; name: string };
+};
+
```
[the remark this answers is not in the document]
BowTiedRadone: No, this is the `implemented trait` type, which is slightly different from the `imported trait` type. The imported type is what we get during AST parameter nodes parsing, the implemented type is a direct entry of a trait implementation contract's AST.

traits.ts
```diff
@@ -0,0 +1,510 @@
...
+} from "./shared.types";
+import { Simnet } from "@hirosystems/clarinet-sdk";
+import { ImplementedTraitType, ImportedTraitType } from "./traits.types";
+
+/**
+ * Enriches contract interface with trait reference data. Before enrichment,
+ * the contract interface lacks trait reference data for parameters. This
+ * function constructs a copy of the contract interface with trait reference
+ * data for parameters that are trait references.
+ * @param ast The contract AST.
```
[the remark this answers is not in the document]
BowTiedRadone: We should not. After closing this, let's tackle this subject in a housekeeping PR 🚀

example/.cache/requirements/SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.sip-010-trait-ft-standard.clar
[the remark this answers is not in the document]
BowTiedRadone: Sure!

invariant.ts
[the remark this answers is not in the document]
BowTiedRadone: We can also do this in the next housekeeping PR! 🚀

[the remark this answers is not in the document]
moodmosaic: Sounds good. Indeed. Then no action needed there.

example/Clarinet.toml
```diff
@@ -6,6 +6,14 @@ cache_dir = "./.cache"
 [[project.requirements]]
 contract_id = 'SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.ststx-token'
 
+[[project.requirements]]
+contract_id = 'SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.sip-010-trait-ft-standard'
+
+[contracts.rendezvous-token]
+path = 'contracts/rendezvous-token.clar'
+clarity_version = 2
+epoch = 2.5
```
moodmosaic, line 15: To make it easier for future contributors, could we add a comment here explaining why this contract has epoch 2.5?

example/Clarinet.toml
```diff
@@ -26,8 +34,8 @@ path = "contracts/slice.clar"
 clarity_version = 3
 epoch = 3.0
 
-[contracts.trait]
-path = 'contracts/trait.clar'
+[contracts.ft-transfer-many]
+path = 'contracts/ft-transfer-many.clar'
```
moodmosaic, line 38: Some paths in this file use double quotes and some others use single quotes. Let's stick with one?

example/contracts/ft-transfer-many.clar
```diff
@@ -0,0 +1,21 @@
+(use-trait ft-trait 'SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.sip-010-trait-ft-standard.sip-010-trait)
+
+(define-public (transfer-many
```
moodmosaic, line 3:
```suggestion
(define-public (send-tokens
```
🤔

example/contracts/ft-transfer-many.clar
```diff
@@ -0,0 +1,21 @@
...
+    (transfers-list
```
moodmosaic, line 4:
```suggestion
    (transfers
```
🤔

example/contracts/ft-transfer-many.clar
```diff
@@ -0,0 +1,21 @@
...
+      (list 5 {token: <ft-trait>, recipient: principal, amount: uint})
```
moodmosaic, line 5:
```suggestion
      (list 5 {token: <ft-trait>, to: principal, amount: uint})
```
🤔

example/contracts/ft-transfer-many.clar
```diff
@@ -0,0 +1,21 @@
...
+    )
+  )
+  (ok (map transfer transfers-list))
```
moodmosaic, line 8:
```suggestion
  (ok (map transfer transfers))
```
🤔

example/contracts/ft-transfer-many.clar
```diff
@@ -0,0 +1,21 @@
...
+)
+
+(define-private (transfer
+    (one-transfer {token: <ft-trait>, recipient: principal, amount: uint})
```
moodmosaic, line 12:
```suggestion
    (one-transfer {token: <ft-trait>, to: principal, amount: uint})
```
🤔

example/Clarinet.toml
```diff
@@ -26,8 +34,8 @@ path = "contracts/slice.clar"
 clarity_version = 3
 epoch = 3.0
 
-[contracts.trait]
-path = 'contracts/trait.clar'
+[contracts.ft-transfer-many]
```
moodmosaic, line 37: Perhaps a more friendly name could be `send-tokens`? We have `counter`, `cargo`, etc. (Like, imagine if `cargo` was called `creade-read-upate-delete.clar`.) 👀

example/contracts/ft-transfer-many.tests.clar
```diff
@@ -0,0 +1,47 @@
+;; Invariants
+
+(define-read-only (invariant-trait (address principal))
```
moodmosaic, line 3: Is `invariant-trait` left over from the previous contract (trait.clar)?

moodmosaic (approved):
https://github.com/stacks-network/rendezvous/pull/88#pullrequestreview-2571841595

PR #90, opened 2025-01-15
Preparation for initial NPM release

### LINKED ISSUE #89: Automate NPM publishing on GitHub Release
Cutting a release off of a tag should trigger a GitHub Action to publish package on NPM registry. Need:

1. Add NPM_TOKEN to repo secrets
2. Example workflow:

```yaml
name: npm-publish
on:
  release:
    types: [created]
jobs:
  publish:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-node@v4
        with:
          node-version: '22'
          registry-url: 'https://registry.npmjs.org'
      - run: npm ci
      - run: npm test
      - run: npm publish
        env:
          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```

Using Node.js 22 LTS for longest support timeline (until October 2026 as per the [release schedule](https://nodejs.org/en/about/previous-releases)).

BowTiedRadone:
This PR prepares `Rendezvous` for the initial NPM publish. It tackles the first part of #89. After the NPM package is initialized and properly configured, we will apply 60db092 via a different PR.


moodmosaic (commented):
>eventually, the name field should be prefixed with the npm scope in package.json

I'd rather do this now, then have @wileyj review and merge.

.github/workflows/npm-publish-release.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+          registry-url: "https://registry.npmjs.org"
+      - run: npm ci
+      - run: npm test
+      - run: npm publish --access public
```
wileyj, line 16: while this would work, is there any reason something like this wasn't used instead?
https://github.com/JS-DevTools/npm-publish

keeping it as-is would require us to manually update the workflow as we want to chagne things (i.e. tagging a release), and also make it easier for a second step that uses the published package to confirm the upload was successful.

BowTiedRadone: Thank you for sharing! Good point. No, just that we extracted this from the [GitHub docs](https://docs.github.com/en/actions/use-cases-and-examples/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry). Will update the workflow 🚀

wileyj: there are some downsides to that approach too though:
- it relies on an outside workflow that needs to be updated. if it's broken or no longer maintained - no publishes happen.

ultimately, either way is fine - i would choose the approach that suits rendezvous better (and it can be changed simply if needed in the future)

moodmosaic: Let’s keep our own and if we see that we need more, perhaps then we switch to the external one.

moodmosaic (approved):
🚢

Issue #97, opened 2025-01-21
Support state transitions for dependent contracts during invariant testing


BowTiedRadone:
When a smart contract depends on another contract (e.g., a function in the target contract calls another contract), include state transitions for all contracts in the dependency chain during invariant testing. This means:

- Identify all contracts directly or indirectly involved in the dependency tree of the target contract.
- Enable calling random public functions with randomly generated arguments for each of these contracts.

This approach ensures state transitions across the entire dependency chain, improving the accuracy of invariant testing.

BowTiedRadone:
Thanks to @bowtiedblox for pointing out this idea!

moodmosaic:
It's an interesting idea but tricky in practice.

Covering every contract in the dependency chain risks combinatorial explosion, as the state space grows rapidly. Mocks and stubs (and Test Doubles in general) may actually help here.

Imagine that the main contract (SUT) manages positions for a CLMM/DEX pool (e.g., a *vault*). It interacts with the pool's range, tick, and math functions, plus other libraries. Targeting the entire dependency chain (pool code, price oracles, etc.) makes the complexity unmanageable.

We likely only care about how the *vault* updates user liquidity or handles rebalances. Test Doubles help simulate relevant pool state changes (e.g., tick moves, fees, liquidity updates), helping detect the unexpected without integrating the full DEX code.

This questioning also applies to example/unit-based and property-based testing: in the context of testing the SUT, is there a use case where touching all its dependencies uncovered a bug missed otherwise?

BowTiedRadone:
@moodmosaic Thank you for the detailed and well-documented comment, as always! Based on the `Test Double` approach, here’s how a stacking pool contract developer can cause state transitions in the dependent contract:  

For example, if a stacking pool contract depends on `pox-4` and is the target contract of a Rendezvous fuzzing session, the solution would be to temporarily move the dependent functions into the SUT (System Under Test). This effectively makes them test doubles:  

```clarity
(contract-call? .pox-4 function-name function-args)  
;; becomes  
(function-name (function-args))  
;; where function-name is temporarily moved to the SUT for fuzz testing purposes  
```  

However, we can still cause state transitions in the boot contracts (including pox-4), as they are responsible for the chain functionality and advancement.

moodmosaic:
> The solution would be to temporarily move the dependent functions into the SUT

[...]

PR #107, opened 2025-01-31
Custom manifest file support

BowTiedRadone:
This PR adds support for custom manifest files for each target contract. To use one, place a file named `Clarinet-<contract-name>.toml` in the target Clarinet project's root.

### Example  
For the command:  
```sh
npx rv sbtc/contracts sbtc-token invariant
```  
Rendezvous will check for `sbtc-token/contracts/Clarinet-sbtc-token.toml`. If found, it will be used to initialize the `Simnet`. If not, the default `Clarinet.toml` will be used.

### Real-World Use Case  
When fuzzing `sbtc-token`, a user may need to replace `sbtc-registry` with a test double. This allows bypassing the `is-protocol-caller` check, enabling state transitions.

#### Original `sbtc-registry.clar`
[See sbtc-registry](https://github.com/stacks-network/sbtc/blob/1b9e1f07f90d4f1f63cc6d91a6d45b33a6febb40/contracts/contracts/sbtc-registry.clar#L360-L369)
```clarity
(define-read-only (is-protocol-caller (contract-flag (buff 1)) (contract principal))
  (begin
    (asserts! (is-eq (some contract) (map-get? active-protocol-contracts contract-flag)) ERR_UNAUTHORIZED)
    (asserts! (is-eq (some contract-flag) (map-get? active-protocol-roles contract)) ERR_UNAUTHORIZED)
    (ok true)
  )
)
```
#### Test Double `sbtc-registry-double.clar`
```clarity
(define-constant deployer tx-sender)

(define-read-only (is-protocol-caller (contract-flag (buff 1)) (contract principal))
  (begin
    (asserts! (is-eq tx-sender deployer) ERR_UNAUTHORIZED)
    (ok true)
  )
)
```

#### Rendezvous Property Test `sbtc-token.tests.clar`
```clarity
(define-public (test-protocol-set-token-uri
    (new-uri (optional (string-utf8 256)))
    (contract-flag (buff 1))
  )
  (let (
      (initial-uri (unwrap-panic (get-token-uri)))
    )
    (ok
      (if
        (or
          (not (is-eq tx-sender deployer))
          (is-eq initial-uri new-uri)
        )
        false
        (begin
          (try! (protocol-set-token-uri new-uri contract-flag))
          (asserts!
            (is-eq
              (unwrap-panic (get-token-uri))
              new-uri
            )
            ERR_FAILED_ASSERTION
          )
          true
        )
      )
    )
  )
)
```
### Custom Manifest Setup  
To use the test double, create a custom manifest file:  
```toml
# Clarinet-sbtc-token.toml
# unchanged...
[contracts.sbtc-registry]
path = 'contracts/sbtc-registry-double.clar'
clarity_version = 3
epoch = 3.0
# unchanged...
```

This PR can represent a solution for #97, where @moodmosaic pointed out the Test Doubles approach in https://github.com/stacks-network/rendezvous/issues/97#issuecomment-2614610694.

### DIFF
--- a/app.tests.ts
+++ b/app.tests.ts
@@ -1,5 +1,5 @@
 import { red } from "ansicolor";
-import { main } from "./app";
+import { getManifestFileName, main } from "./app";
 import { version } from "./package.json";
 
 describe("Command-line arguments handling", () => {
@@ -365,4 +365,18 @@ describe("Command-line arguments handling", () => {
       jest.restoreAllMocks();
     }
   );
+});
+
+describe("Custom manifest detection", () => {
+  it("returns the default manifest file name for the example project", () => {
+    // Arrange
+    const manifestDir = "example";
+    const targetContractName = "counter";
+
+    // Act
+    const actual = getManifestFileName(manifestDir, targetContractName);
+
+    // Assert
+    expect(actual).toBe("Clarinet.toml");
+  });
 });
--- a/app.tests.ts
+++ b/app.tests.ts
@@ -1,9 +1,8 @@
 import { red } from "ansicolor";
 import { getManifestFileName, main } from "./app";
 import { version } from "./package.json";
-import { cpSync, mkdtempSync, rmdirSync, writeFileSync } from "fs";
-import { join, resolve } from "path";
-import { tmpdir } from "os";
+import { resolve } from "path";
+import fs from "fs";
 
 describe("Command-line arguments handling", () => {
   const initialArgv = process.argv;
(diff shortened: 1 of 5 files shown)


app.tests.ts
```diff
@@ -366,3 +366,17 @@ describe("Command-line arguments handling", () => {
     }
   );
 });
+
+describe("Custom manifest detection", () => {
```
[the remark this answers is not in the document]
BowTiedRadone: A test for the custom manifest file name detection was added in 342c3a6.

app.tests.ts
```diff
@@ -1,6 +1,9 @@
 import { red } from "ansicolor";
 import { getManifestFileName, main } from "./app";
 import { version } from "./package.json";
+import { cpSync, mkdtempSync, rmdirSync, writeFileSync } from "fs";
+import { join, resolve } from "path";
+import { tmpdir } from "os";
```
moodmosaic, line 6: Those (many) imports usually hint that we could instead use a test double (mock or stub).

app.tests.ts
```diff
@@ -379,4 +382,44 @@ describe("Custom manifest detection", () => {
...
+    const exampleProject = "example";
+    const tempDir = mkdtempSync(join(tmpdir(), "simnet-test-"));
+
+    // Create a custom manifest file complying with the Rendezvous convention:
+    // `Clarinet-<target-contract-name>.toml`.
+    const expected = "Clarinet-counter.toml";
+
+    cpSync(exampleProject, tempDir, { recursive: true });
+
+    writeFileSync(
```
moodmosaic, line 397: Couldn't we just use a (strict) mock here? I think it'll require less imports and we would also avoid touching the underlying fs (wrote in Sublime Text, might need to be tweaked to compile):

```ts
// Arrange
const manifestDir =
  "d290f1ee-6c54-4b01-90e6-d701748f0851";
const target =
  "7c9e6679-7425-40de-944b-e07fc1f90ae7.clar";

const expected = `Clarinet-${target}.toml`;
const expectedPath = resolve(manifestDir, expected);

const mock = jest
  .spyOn(fs, "existsSync")
  .mockImplementation((p: string) => p === expectedPath);

// Act
const actual = getManifestFileName(manifestDir, target);

// Assert
expect(actual).toBe(expected);
```

BowTiedRadone: Great suggestion! Thank you 🙏

moodmosaic (approved):

Issue #86, opened 2025-01-09
Support graceful termination via CTRL+C


moodmosaic:
Currently CTRL+C doesn't cleanly terminate test sessions. Should:
- Stop any running test immediately
- Print results of completed tests
- Show partial coverage data if available
- Clean up any temporary state

Example:
```
Starting invariant testing type for the counter contract...

₿        1 Ӿ        9   wallet_4        counter decrement 
₿        1 Ӿ        9   wallet_4 [PASS] counter invariant-counter-gt-zero 
₿      835 Ӿ      845   wallet_7        counter add 6
₿      835 Ӿ      845   wallet_1 [PASS] counter invariant-counter-gt-zero 
^C
Terminating...
Tests:      4 passed, 2 discarded, 94 skipped
Time:       1.24s
Coverage:   7.5%
```

This requires #85 as prerequisite.

PR #109, opened 2025-02-04
Add support for `dialers` in invariant testing

### LINKED ISSUE #59: Add printed events to the context for assertions
When calling `callPublicFn`, we should add any events to the Rendezvous context.  

I haven't fully thought about this yet, but I'm opening this to support what @friedger worked on in https://github.com/stacks-network/sbtc/pull/1091.

This also opens the door for adding support for *templates* (or *dealers* in the Rendezvous DSL), which I'll track as a separate issue. Using *dealers*, it would be possible to run `rv` against a Clarinet project and have it check all its SIP-10 contracts, for example.

BowTiedRadone:
This PR introduces support for custom `pre-` and `post-execution` JavaScript functions, called `dialers`, in invariant testing. Dialers run before and after public function calls, allowing users to add custom assertions, behaviors, or side effects during execution. Thanks to @moodmosaic for designing this feature in https://github.com/stacks-network/rendezvous/issues/59#issuecomment-2626703488. Resolves #59.

### DIFF
--- a/app.tests.ts
+++ b/app.tests.ts
@@ -10,7 +10,7 @@ describe("Command-line arguments handling", () => {
   const helpMessage = `
   rv v${version}
   
-  Usage: rv <path-to-clarinet-project> <contract-name> <type> [--seed=<seed>] [--path=<path>] [--runs=<runs>]
+  Usage: rv <path-to-clarinet-project> <contract-name> <type> [--seed=<seed>] [--path=<path>] [--runs=<runs>] [--dial=<path-to-dialers-file>]
 
   Positional arguments:
     path-to-clarinet-project - The path to the Clarinet project.
@@ -21,6 +21,7 @@ describe("Command-line arguments handling", () => {
     --seed - The seed to use for the replay functionality.
     --path - The path to use for the replay functionality.
     --runs - The runs to use for iterating over the tests. Default: 100.
+    --dial – The path to a JavaScript file containing custom pre- and post-execution functions (dialers).
     --help - Show the help message.
   `;
   const noManifestMessage = red(
--- a/app.tests.ts
+++ b/app.tests.ts
@@ -10,7 +10,7 @@ describe("Command-line arguments handling", () => {
   const helpMessage = `
   rv v${version}
   
-  Usage: rv <path-to-clarinet-project> <contract-name> <type> [--seed=<seed>] [--path=<path>] [--runs=<runs>] [--dial=<path-to-dialers-file>]
+  Usage: rv <path-to-clarinet-project> <contract-name> <type> [--seed=<seed>] [--path=<path>] [--runs=<runs>] [--dial=<path-to-dialers-file>] [--help]
 
   Positional arguments:
     path-to-clarinet-project - The path to the Clarinet project.
--- a/app.tests.ts
+++ b/app.tests.ts
@@ -197,6 +197,23 @@ describe("Command-line arguments handling", () => {
       ],
     ],
     [
+      ["manifest path", "contract name", "type=invariant", "dialers file path"],
+      [
+        "node",
+        "app.js",
+        "example",
+        "counter",
+        "invariant",
+        "--dial=example/sip010.js",
+      ],
+      [
+        `Using manifest path: example/Clarinet.toml`,
+        `Target contract: counter`,
+        `Using dial path: example/sip010.js`,
+        `\nStarting invariant testing type for the counter contract...\n`,
+      ],
+    ],
+    [
       ["manifest path", "contract name", "type=test"],
       ["node", "app.js", "example", "counter", "test"],
       [
--- a/app.ts
+++ b/app.ts
@@ -12,6 +12,7 @@ import { issueFirstClassCitizenship } from "./citizen";
 import { version } from "./package.json";
 import { red } from "ansicolor";
 import { existsSync } from "fs";
+import { DialerRegistry } from "./dialer";
 
 const logger = (log: string, logLevel: "log" | "error" | "info" = "log") => {
   console[logLevel](log);
@@ -140,6 +141,27 @@ export async function main() {
     radio.emit("logMessage", `Using runs: ${runs}`);
   }
 
+  /**
+   * The path to the dialer file. The dialer file allows the user to register
+   * custom pre and post-execution JavaScript functions to be executed before
+   * and after the public function calls during invariant testing.
+   */
+  const dialPath = parseOptionalArgument("dial") || undefined;
+  if (dialPath !== undefined) {
+    radio.emit("logMessage", `Using dial path: ${dialPath}`);
+  }
+
+  /**
+   * The dialer registry, which is used to keep track of all the custom dialers
+   * registered by the user using the `--dial` flag.
+   */
+  const dialerRegistry =
+    dialPath !== undefined ? new DialerRegistry(dialPath) : undefined;
+
+  if (dialerRegistry !== undefined) {
+    dialerRegistry.registerDialers();
+  }
+
   const simnet = await issueFirstClassCitizenship(
     manifestDir,
     manifestPath,
@@ -168,7 +190,7 @@ export async function main() {
   // If "test", call `checkProperties` for property-based testing.
   switch (type) {
     case "invariant": {
-      checkInvariants(
+      await checkInvariants(
         simnet,
         sutContractName,
         rendezvousList,
@@ -176,6 +198,7 @@ export async function main() {
         seed,
         path,
         runs,
+        dialerRegistry,
         radio
       );
       break;
--- a/app.ts
+++ b/app.ts
@@ -44,7 +44,7 @@ export const getManifestFileName = (
 const helpMessage = `
   rv v${version}
   
-  Usage: rv <path-to-clarinet-project> <contract-name> <type> [--seed=<seed>] [--path=<path>] [--runs=<runs>]
+  Usage: rv <path-to-clarinet-project> <contract-name> <type> [--seed=<seed>] [--path=<path>] [--runs=<runs>] [--dial=<path-to-dialers-file>]
 
   Positional arguments:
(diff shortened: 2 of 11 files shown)


dialer.ts
```diff
@@ -0,0 +1,65 @@
...
+      return;
+    }
+
+    for (const dial of this.postDialers) {
+      await dial(context);
+    }
+  }
+}
+
+export const dialerRegistry = new DialerRegistry();
```
moodmosaic, line 39: `DialerRegistry` should take a `path` through the constructor and instantiated in the root, app.ts.

dialer.ts
```diff
@@ -0,0 +1,65 @@
...
+        if (key.startsWith("pre")) {
+          dialerRegistry.registerPreDialer(fn as Dialer);
+        } else if (key.startsWith("post")) {
+          dialerRegistry.registerPostDialer(fn as Dialer);
+        }
+      }
+    });
+  } catch (error) {
+    console.error(`Failed to load dialers:`, error);
+    process.exit(1);
```
moodmosaic, line 63: Not sure if it should exit `rv`. Perhaps `rv` should run without calling any of the registered functions.

app.ts
```diff
@@ -134,15 +144,16 @@ export async function main() {
   // If "test", call `checkProperties` for property-based testing.
   switch (type) {
     case "invariant": {
-      checkInvariants(
+      await checkInvariants(
```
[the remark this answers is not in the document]
BowTiedRadone: Since `checkProperties` is not yet asynchronous, I don't think the `await` is really needed here. We can add it as soon as/if the function becomes asynchronous.

BowTiedRadone:
Before merging this PR, we should update the `README` to reflect the final `dialers` feature.

invariant.ts
```diff
@@ -314,22 +340,31 @@ export const checkInvariants = (
...
-              dim(
-                `₿ ${simnet.burnBlockHeight.toString().padStart(8)} ` +
-                  `Ӿ ${simnet.blockHeight.toString().padStart(8)}   ` +
-                  `${sutCallerWallet}        ` +
-                  `${targetContractName} ` +
-                  `${underline(selectedFunction.name)} ` +
-                  printedFunctionArgs
-              )
-            );
+            // If the pre dialer fails, throw an error.
```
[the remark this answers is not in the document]
moodmosaic: As discussed, this is required so that a failing pre- or post- dialer can cause the loop to end and report the error.

invariant.ts
```diff
@@ -300,6 +316,16 @@ export const checkInvariants = (
...
               );
+
+              try {
+                await dialerRegistry.executePostDialers({
+                  selectedFunction: selectedFunction,
+                  functionCall: functionCall,
+                  clarityValueArguments: selectedFunctionsArgsCV[index],
+                });
+              } catch (e: any) {
+                throw new Error(`Error in post dialer: ${e.message}`);
```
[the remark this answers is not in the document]
moodmosaic: As discussed, this is required so that a failing pre- or post- dialer can cause the loop to end and report the error. However, in the re-review that is about to start we might revisit this more thoroughly.

dialer.ts
moodmosaic: This a good point. Agreed. Let's make sure it can work well with #86. (I think it will, as it won't be even the case.)

example/sip010.js
```diff
@@ -0,0 +1,73 @@
+// This file contains a custom predefined post-dialer that checks the SIP-010
+// transfer function for the print event. For a token to be SIP-010 compliant,
+// its transfer function must emit the print event with the `memo` content.
+//
+// This is just an example of the power of custom dialers. In sBTC, the
+// transfer function did not emit the print event, resulting in sBTC not being
+// SIP-010 compliant. This custom dialer allows for the detection of such
```
moodmosaic, line 7:
```suggestion
// SIP-010 compliant: https://github.com/stacks-network/sbtc/issues/1090.
//
// This custom dialer allows for the detection of such
```

dialer.types.ts
```diff
@@ -0,0 +1,11 @@
+import { ParsedTransactionResult } from "@hirosystems/clarinet-sdk";
+import { ClarityValue } from "@stacks/transactions";
+import { EnrichedContractInterfaceFunction } from "./shared.types";
+
+export type Dialer = (context: DialerContext) => Promise<void> | void;
+
+type DialerContext = {
```
moodmosaic, line 7:
```suggestion
export type DialerContext = {
```

dialer.types.ts
```diff
@@ -0,0 +1,11 @@
...
+  clarityValueArguments: ClarityValue[];
+  functionCall: ParsedTransactionResult;
```
moodmosaic, line 9:
```suggestion
  functionCall: ParsedTransactionResult | undefined;
```

moodmosaic: (Needs the `undefined` here because in the `pre-` phase this is undefined.)

dialer.ts
```diff
@@ -0,0 +1,67 @@
+import { existsSync } from "fs";
+import { resolve } from "path";
+import { Dialer } from "./dialer.types";
```
moodmosaic, line 3:
```suggestion
import { Dialer, DialerContext } from "./dialer.types";
```

dialer.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export class DialerRegistry {
+  private dialPath: string;
```
moodmosaic, line 6:
```suggestion
  private readonly dialPath: string;
```

dialer.ts
```diff
@@ -0,0 +1,67 @@
...
+          }
+        }
+      });
+    } catch (error) {
+      console.error(`Failed to load dialers:`, error);
+      process.exit(1);
+    }
+  }
+
+  async executePreDialers(context: any) {
```
moodmosaic, line 48:
```suggestion
  async executePreDialers(context: DialerContext) {
```

dialer.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (this.preDialers.length === 0) {
+      return;
+    }
+
+    for (const dial of this.preDialers) {
+      await dial(context);
+    }
+  }
+
+  async executePostDialers(context: any) {
```
moodmosaic, line 58:
```suggestion
  async executePostDialers(context: DialerContext) {
```

dialer.tests.ts
```diff
@@ -0,0 +1,87 @@
...
+    const actual = new DialerRegistry(dialPath);
+
+    // Assert
+    expect(actual).toBeInstanceOf(DialerRegistry);
+  });
+
+  it("correctly executes registered pre-dialer", async () => {
+    // Arrange
+    const mockPreDialer = jest.fn();
+    const dialerRegistry = new DialerRegistry(dialPath);
```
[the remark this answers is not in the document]
moodmosaic: Changed to `registry` to be consistent with the other tests below.

dialer.tests.ts
```diff
@@ -0,0 +1,87 @@
...
+    // Arrange
+    const mockPreDialer1 = jest.fn();
+    const mockPreDialer2 = jest.fn();
+    const registry = new DialerRegistry(dialPath);
+
+    registry.registerPreDialer(mockPreDialer1);
+    registry.registerPreDialer(mockPreDialer2);
+
+    // Act
+    await registry.executePreDialers({});
```
moodmosaic, line 51:
```suggestion
    await registry.executePreDialers({} as any as DialerContext);
```

dialer.tests.ts
```diff
@@ -0,0 +1,87 @@
...
+    // Arrange
+    const mockPostDialer1 = jest.fn();
+    const mockPostDialer2 = jest.fn();
+    const registry = new DialerRegistry(dialPath);
+
+    registry.registerPostDialer(mockPostDialer1);
+    registry.registerPostDialer(mockPostDialer2);
+
+    // Act
+    await registry.executePostDialers({});
```
moodmosaic, line 68:
```suggestion
    await registry.executePostDialers({} as any as DialerContext);
```

moodmosaic (commented):
Very nice! Left some comments mostly around using the context type instead of any type.

dialer.types.ts
BowTiedRadone: Exactly 🚀

app.tests.ts
```diff
@@ -196,6 +196,23 @@ describe("Command-line arguments handling", () => {
...
     ],
+    [
+      ["manifest path", "contract name", "type=invariant", "dialers file path"],
+      [
+        "node",
+        "app.js",
+        "example",
+        "counter",
+        "invariant",
+        "--dial=example/sip010.js",
```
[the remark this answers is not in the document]
BowTiedRadone: Great catch! This will be refactored in a follow-up PR. 🚀

invariant.ts
```diff
@@ -314,22 +341,31 @@ export const checkInvariants = (
...
-                `₿ ${simnet.burnBlockHeight.toString().padStart(8)} ` +
-                  `Ӿ ${simnet.blockHeight.toString().padStart(8)}   ` +
-                  `${sutCallerWallet}        ` +
-                  `${targetContractName} ` +
-                  `${underline(selectedFunction.name)} ` +
-                  printedFunctionArgs
-              )
-            );
+            // If the pre dialer fails, throw an error.
+            if (error.message.startsWith("Error in pre dialer")) {
```
moodmosaic, line 345: This can be fragile if a single character or casing changes in this message. Thoughts?

invariant.ts
```diff
@@ -314,22 +341,31 @@ export const checkInvariants = (
...
+              throw error;
+            }
+            // If the post dialer fails, throw an error.
+            if (error.message.startsWith("Error in post dialer")) {
```
moodmosaic, line 349: This can be fragile if a single character or casing changes in this message. Thoughts?

moodmosaic (approved):
LGTM! 🚀

PR #111, opened 2025-02-07
Fix first-class citizenship simnet wallet balances

BowTiedRadone:
This PR fixes a bug in `issueFirstClassCitizenship`. The `initEmptySession` method did not set wallet balances from `Devnet.toml`, causing all simnet accounts to start with 0 STX. This fix ensures that after initializing a new session, the correct STX balances are minted as specified in `Devnet.toml`.

### DIFF
--- a/citizen.tests.ts
+++ b/citizen.tests.ts
@@ -13,6 +13,7 @@ import { cpSync, existsSync, mkdtempSync, readFileSync, rmSync } from "fs";
 import { tmpdir } from "os";
 import yaml from "yaml";
 import { getManifestFileName } from "./app";
+import { cvToValue, hexToCV } from "@stacks/transactions";
 
 describe("Simnet deployment plan operations", () => {
   const manifestDir = "example";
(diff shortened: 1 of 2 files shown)


citizen.ts
```diff
@@ -44,8 +45,27 @@ export const issueFirstClassCitizenship = async (
...
+      } catch (error) {
+        return [address, BigInt(0)];
+      }
+    })
+  );
+
   await simnet.initEmptySession();
 
+  // The `initEmptySession` method will reset the balances of the accounts to
+  // 0. Restore the balances to their original values.
```
hugocaillard, line 64: This comment can be misleading.
`initEmptySession` **does not** reset the balances.
But, it's an "empty shell" that doesn't read the Clarinet.toml or Devnet.toml at all.
So it does not come with any initial founded accounts

moodmosaic: Thanks @hugocaillard! 👍 Yeah, we don't even need this comment—the session is empty, so we have to send the TXs manually, just like @BowTiedRadone did.  

Good catch @BowTiedRadone! Easy to miss when switching to `initEmptySession`, though this is actually the correct behavior from the SDK’s perspective. I missed it too when reviewing.

moodmosaic, line 64:
```suggestion
```

citizen.ts
```diff
@@ -44,8 +45,27 @@ export const issueFirstClassCitizenship = async (
...
 
+  const simnetAddresses = [...simnet.getAccounts().values()];
+
+  const balancesMap = new Map(
+    Array.from(simnetAddresses, (address) => {
+      try {
+        const balanceHex = simnet.runSnippet(`(stx-get-balance '${address})`);
+        return [address, cvToValue(hexToCV(balanceHex))];
+      } catch (error) {
+        return [address, BigInt(0)];
```
moodmosaic, line 56: I'm not sure we should catch this. It should propagate up. Can we do anything meaningful with a tuple of `address, BigInt(0)`?

citizen.tests.ts
```diff
@@ -359,4 +360,38 @@ describe("Simnet deployment plan operations", () => {
...
+    // Verify
+    const balancesMap = new Map(
+      Array.from([...firstClassSimnet.getAccounts().values()], (address) => {
+        try {
+          const balanceHex = firstClassSimnet.runSnippet(
+            `(stx-get-balance '${address})`
+          );
+          return [address, cvToValue(hexToCV(balanceHex))];
+        } catch (error) {
+          return [address, BigInt(0)];
```
moodmosaic, line 385: Do we hit this `catch` block ever?

citizen.tests.ts
```diff
@@ -359,4 +360,38 @@ describe("Simnet deployment plan operations", () => {
     // Teardown
     rmSync(tempDir, { recursive: true, force: true });
   });
+
+  it(`the first-class citizenship simnet has the correct balances for the registered accounts`, async () => {
+    // Setup
+    const tempDir = mkdtempSync(join(tmpdir(), "simnet-test-"));
```
moodmosaic, line 366:
```suggestion
    const tempDir = mkdtempSync(join(tmpdir(), "simnet-test"));
```
Do we need the last dash?

moodmosaic (commented):
Good catch! Users would definitely expect `rv` to respect their `.toml` file. Nice spot, @BowTiedRadone! Thanks for checking as well, @hugocaillard.

🚀

citizen.tests.ts
BowTiedRadone: You are right. Will remove the try-catch. Good catch!

moodmosaic (approved):
LGTM!

PR #113, opened 2025-02-07
Display the error code for all `[FAIL]` logs in property testing

### LINKED ISSUE #112: Show error codes in failure logs during property testing shrinking
Successful calls during shrinking might hide the initial failure. Users need to see both the first and subsequent failures during this process. We can display them at the end of a `[FAIL]` log like this: `(err <error-code>)`.

**Example Output:**
```
Starting property testing type for the counter contract...

₿        5 Ӿ       11   deployer [PASS] counter test-add 2023153563
₿        5 Ӿ       12   wallet_2 [FAIL] counter test-increment  (err 404)
₿        5 Ӿ       13   wallet_4 [FAIL] counter test-increment  (err 404)
₿        5 Ӿ       14   wallet_4 [PASS] counter test-add 1188522168
```

BowTiedRadone:
This PR updates the failure logs in property testing to include the failure error code. Resolves #112.

### DIFF
--- a/property.ts
+++ b/property.ts
@@ -319,8 +319,10 @@ export const checkProperties = (
           } catch (error: any) {
             const displayedError =
               error instanceof PropertyTestError
-                ? error.errorCode
-                : error.includes("Runtime")
+                ? error.clarityError
+                : error &&
+                  typeof error === "string" &&
+                  error.toLowerCase().includes("runtime")
                 ? "(runtime)"
                 : "(unknown)";
 
@@ -487,9 +489,9 @@ export const isReturnTypeBoolean = (
 ) => discardFunctionInterface.outputs.type === "bool";
 
 class PropertyTestError extends Error {
-  readonly errorCode: any;
-  constructor(message: string, errorCode: string) {
+  readonly clarityError: string;
+  constructor(message: string, clarityError: string) {
     super(message);
-    this.errorCode = errorCode;
+    this.clarityError = clarityError;
   }
 }
--- a/property.ts
+++ b/property.ts
@@ -313,7 +313,9 @@ export const checkProperties = (
             }
           } catch (error: any) {
             const displayedError =
-              error instanceof PropertyTestError ? `(err ${error.code})` : "";
+              error instanceof PropertyTestError
+                ? `(err ${error.errorCode})`
+                : "";
             // Capture the error and log the test failure.
             radio.emit(
               "logMessage",
@@ -477,9 +479,9 @@ export const isReturnTypeBoolean = (
 ) => discardFunctionInterface.outputs.type === "bool";
 
 class PropertyTestError extends Error {
-  code: any;
+  readonly errorCode: any;
   constructor(message: string, errorCode: any) {
     super(message);
-    this.code = errorCode;
+    this.errorCode = errorCode;
   }
 }
--- a/property.ts
+++ b/property.ts
@@ -306,11 +306,14 @@ export const checkProperties = (
                 simnet.mineEmptyBurnBlocks(r.burnBlocks);
               }
             } else {
-              throw new Error(
-                `Test failed for ${targetContractName} contract: "${r.selectedTestFunction.name}" returned ${testFunctionCallResultJson.value.value}`
+              throw new PropertyTestError(
+                `Test failed for ${targetContractName} contract: "${r.selectedTestFunction.name}" returned ${testFunctionCallResultJson.value.value}`,
+                testFunctionCallResultJson.value.value
               );
             }
           } catch (error: any) {
+            const displayedError =
+              error instanceof PropertyTestError ? `(err ${error.code})` : "";
             // Capture the error and log the test failure.
             radio.emit(
               "logMessage",
@@ -321,7 +324,8 @@ export const checkProperties = (
                   `[FAIL] ` +
                   `${targetContractName} ` +
                   `${underline(r.selectedTestFunction.name)} ` +
-                  printedTestFunctionArgs
+                  `${printedTestFunctionArgs} ` +
+                  displayedError
               )
             );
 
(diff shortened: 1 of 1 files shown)


property.ts
```diff
@@ -306,11 +306,14 @@ export const checkProperties = (
                 simnet.mineEmptyBurnBlocks(r.burnBlocks);
               }
             } else {
-              throw new Error(
-                `Test failed for ${targetContractName} contract: "${r.selectedTestFunction.name}" returned ${testFunctionCallResultJson.value.value}`
+              throw new PropertyTestError(
+                `Test failed for ${targetContractName} contract: "${r.selectedTestFunction.name}" returned ${testFunctionCallResultJson.value.value}`,
+                testFunctionCallResultJson.value.value
```
moodmosaic, line 311: Can `testFunctionCallResultJson.value` be null and thus `testFunctionCallResultJson.value.value` throw a null reference error?

property.ts
```diff
@@ -306,11 +306,14 @@ export const checkProperties = (
...
               );
             }
           } catch (error: any) {
+            const displayedError =
+              error instanceof PropertyTestError ? `(err ${error.code})` : "";
```
moodmosaic, line 316: Nice use of custom error types :)

property.ts
```diff
@@ -471,3 +475,11 @@ export const isParamsMatch = (
 export const isReturnTypeBoolean = (
   discardFunctionInterface: ContractInterfaceFunction
 ) => discardFunctionInterface.outputs.type === "bool";
+
+class PropertyTestError extends Error {
+  code: any;
```
moodmosaic, line 480:
```suggestion
  private readonly errorCode: string;
```

property.ts
```diff
@@ -471,3 +475,11 @@ export const isParamsMatch = (
...
+  constructor(message: string, errorCode: any) {
+    super(message);
+    this.code = errorCode;
```
moodmosaic, line 483:
```suggestion
    this.errorCode = errorCode;
```

property.ts
```diff
@@ -471,3 +475,11 @@ export const isParamsMatch = (
 export const isReturnTypeBoolean = (
   discardFunctionInterface: ContractInterfaceFunction
 ) => discardFunctionInterface.outputs.type === "bool";
+
+class PropertyTestError extends Error {
+  code: any;
+  constructor(message: string, errorCode: any) {
```
moodmosaic, line 481:
```suggestion
  constructor(message: string, errorCode: string) {
```

moodmosaic (commented):
🚀

property.ts
BowTiedRadone: I think `readonly` is the needed restriction for the code field. Otherwise, without a getter, we won't be able to use it to display it.

property.ts
BowTiedRadone: No, a valid public function always returns a response type, which, AFAIK, will always have a nested `value` under the first `value` field.

property.ts
BowTiedRadone: 🤺

property.ts
BowTiedRadone: Here, the `errorCode` really is  `any`. It can be an object, a number, a string, etc.

property.ts
BowTiedRadone: Renamed and turned to `readonly` in 74d4c5319dbaea36640493de6aaf87e56573c94f.

property.ts
BowTiedRadone: I just thought about this one more time. It can indeed result in a runtime error.

property.ts
moodmosaic: What if we declare it as `errorCode: string | number`? (What is the case of object?)

property.ts
moodmosaic: Yep, can't we use optional chaining (`?.`) and nullish coalescing (`??`)? E.g.:

```js
testFunctionCallResultJson?.value?.value ?? "unknown"
```

BowTiedRadone: Yep, something like that should work 🚀

property.ts
BowTiedRadone: It can be an object if the type nested under the `response error` is complex (list, tuple, response, etc.)

BowTiedRadone: I think it will be best to take any and try [cvToString](https://github.com/hirosystems/stacks.js/blob/4dcad2cf88375bfedf288b317e4a10cfcf0ddd48/packages/transactions/src/clarity/clarityValue.ts#L38). This will return Clarity syntax e.g. `(err u999)`.

property.ts
```diff
@@ -471,3 +485,11 @@ export const isParamsMatch = (
 export const isReturnTypeBoolean = (
   discardFunctionInterface: ContractInterfaceFunction
 ) => discardFunctionInterface.outputs.type === "bool";
+
+class PropertyTestError extends Error {
+  readonly errorCode: any;
+  constructor(message: string, errorCode: string) {
```
moodmosaic, line 491: Can we come up with a more descriptive name for `errorCode`?

property.ts
```diff
@@ -306,11 +306,24 @@ export const checkProperties = (
...
+              throw new PropertyTestError(
+                `Test failed for ${targetContractName} contract: "${r.selectedTestFunction.name}" returned ${displayedError}`,
+                displayedError
               );
             }
           } catch (error: any) {
+            const displayedError =
+              error instanceof PropertyTestError
+                ? error.errorCode
+                : error.includes("Runtime")
```
moodmosaic, line 323: Perhaps we should lowercase the error first, to make the comparison more robust.

However, what is the advantage of printing “(runtime)” and “(unknown)”? Perhaps in this case nothing should be printed?

property.ts
```diff
@@ -471,3 +485,11 @@ export const isParamsMatch = (
 export const isReturnTypeBoolean = (
   discardFunctionInterface: ContractInterfaceFunction
 ) => discardFunctionInterface.outputs.type === "bool";
+
+class PropertyTestError extends Error {
+  readonly errorCode: any;
```
moodmosaic, line 490: This is probably leftover here with the type `any`.

property.ts
BowTiedRadone: It would be great to see that the error was actually a runtime error. For the unknown, it should not happen if we are not missing something.

property.ts
BowTiedRadone: That is correct. Great catch 🤺

PR #115, opened 2025-02-18
[DRAFT] Update `clarinet-sdk` and `clarinet-sdk-wasm` to the latest version

BowTiedRadone:
This PR updates `clarinet-sdk` and `clarinet-sdk-wasm` to the latest stable version. Before merging the PR, Rendezvous compatibility with the **_remote data_** feature introduced in `@hirosystems/clarinet-sdk v2.13.0` has to be checked. For this reason, this PR is still a draft.

### DIFF
--- a/citizen.ts
+++ b/citizen.ts
@@ -54,7 +54,11 @@ export const issueFirstClassCitizenship = async (
     })
   );
 
-  await simnet.initEmptySession();
+  await simnet.initEmptySession({
+    enabled: false,
+    initial_height: 5,
+    api_url: "",
+  });
 
   simnetAddresses.forEach((address) => {
     simnet.mintSTX(address, balancesMap.get(address)!);
--- a/package.json
+++ b/package.json
@@ -30,14 +30,14 @@
     "url": "https://github.com/stacks-network/rendezvous.git"
   },
   "dependencies": {
-    "@hirosystems/clarinet-sdk": "2.12.0",
+    "@hirosystems/clarinet-sdk": "2.13.0",
     "@stacks/transactions": "^6.16.1",
     "ansicolor": "^2.0.3",
     "fast-check": "^3.20.0",
     "yaml": "^2.6.1"
   },
   "devDependencies": {
-    "@hirosystems/clarinet-sdk-wasm": "2.12.0",
+    "@hirosystems/clarinet-sdk-wasm": "2.13.0",
     "@types/jest": "^29.5.12",
     "jest": "^29.7.0",
     "ts-jest": "^29.2.5",
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.4.1",
       "license": "GPL-3.0-only",
       "dependencies": {
-        "@hirosystems/clarinet-sdk": "2.12.0",
+        "@hirosystems/clarinet-sdk": "2.13.0",
         "@stacks/transactions": "^6.16.1",
         "ansicolor": "^2.0.3",
         "fast-check": "^3.20.0",
@@ -19,7 +19,7 @@
         "rv": "dist/app.js"
       },
       "devDependencies": {
-        "@hirosystems/clarinet-sdk-wasm": "2.12.0",
+        "@hirosystems/clarinet-sdk-wasm": "2.13.0",
         "@types/jest": "^29.5.12",
         "jest": "^29.7.0",
         "ts-jest": "^29.2.5",
@@ -507,12 +507,12 @@
       }
     },
     "node_modules/@hirosystems/clarinet-sdk": {
-      "version": "2.12.0",
-      "resolved": "https://registry.npmjs.org/@hirosystems/clarinet-sdk/-/clarinet-sdk-2.12.0.tgz",
-      "integrity": "sha512-6Jr/THJg4t52QO19ZKVEvIz99pTpSUOGMe1/TeczgzzQnb1k6MORvogWT2UhCGjSyku0VlnYQZS6CBP2xK2XsA==",
+      "version": "2.13.0",
+      "resolved": "https://registry.npmjs.org/@hirosystems/clarinet-sdk/-/clarinet-sdk-2.13.0.tgz",
+      "integrity": "sha512-z6ksxvFWohlur27sv7Wg10R6dt6c8yTY60uLyJCx8NwLx//3nxAb3Uta+oQCDjDh4LxHfsynV7SR92HuCjkFwg==",
       "license": "GPL-3.0",
       "dependencies": {
-        "@hirosystems/clarinet-sdk-wasm": "^2.12.0",
+        "@hirosystems/clarinet-sdk-wasm": "2.13.0",
         "@stacks/transactions": "^6.13.0",
         "kolorist": "^1.8.0",
         "prompts": "^2.4.2",
@@ -527,10 +527,13 @@
       }
     },
     "node_modules/@hirosystems/clarinet-sdk-wasm": {
-      "version": "2.12.0",
-      "resolved": "https://registry.npmjs.org/@hirosystems/clarinet-sdk-wasm/-/clarinet-sdk-wasm-2.12.0.tgz",
-      "integrity": "sha512-82fuVgGZ2pt7fdHc1gTZJK1ryS2LvhWUGDyJXS16j7b5Bge9z0k/qwE2MWcZ5VBYVgbaKLvnaxH1o4Yow05QAQ==",
-      "license": "GPL-3.0"
+      "version": "2.13.0",
+      "resolved": "https://registry.npmjs.org/@hirosystems/clarinet-sdk-wasm/-/clarinet-sdk-wasm-2.13.0.tgz",
+      "integrity": "sha512-AmXYx41uMn9MPTJzXY7fsXUY5E6YJqMYTdQU/6N3wBwms390p6UmxTq5z9CCFalUv4UfYH5DIvePqPLSVI9EHQ==",
+      "license": "GPL-3.0",
+      "dependencies": {
+        "sync-request": "6.1.0"
+      }
     },
     "node_modules/@istanbuljs/load-nyc-config": {
       "version": "1.1.0",
@@ -1023,10 +1026,28 @@
         "@types/node": "*"
       }
     },
+    "node_modules/@types/concat-stream": {
+      "version": "1.6.1",
+      "resolved": "https://registry.npmjs.org/@types/concat-stream/-/concat-stream-1.6.1.tgz",
+      "integrity": "sha512-eHE4cQPoj6ngxBZMvVf6Hw7Mh4jMW4U9lpGmS5GBPB9RYxlFg+CHaVN7ErNY4W9XfLIEn20b4VDYaIrbq0q4uA==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
     "node_modules/@types/estree": {
       "version": "1.0.6",
       "resolved": "https://registry.npmjs.org/@types/estree/-/estree-1.0.6.tgz",
       "integrity": "sha512-AYnb1nQyY49te+VRAVgmzfcgjYS91mY5P0TKUDCLEM+gNnA+3T6rWITXRLYCpahpqSQbN5cE+gHpnPyXjHWxcw=="
+    },
+    "node_modules/@types/form-data": {
+      "version": "0.0.33",
+      "resolved": "https://registry.npmjs.org/@types/form-data/-/form-data-0.0.33.tgz",
+      "integrity": "sha512-8BSvG1kGm83cyJITQMZSulnl6QV8jqAGreJsc5tPu1Jq0vTSOiY/k24Wx82JRpWwZSqrala6sd5rWi6aNXvqcw==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/node": "*"
+      }
     },
     "node_modules/@types/graceful-fs": {
       "version": "4.1.9",
@@ -1078,6 +1099,12 @@
       "dependencies": {
         "undici-types": "~5.26.4"
       }
+    },
+    "node_modules/@types/qs": {
+      "version": "6.9.18",
+      "resolved": "https://registry.npmjs.org/@types/qs/-/qs-6.9.18.tgz",
+      "integrity": "sha512-kK7dgTYDyGqS+e2Q4aK9X3D7q234CIZ1Bv0q/7Z5IwRDoADNU81xXJK/YVyLbLTZCoIwUoDoffFeF+p/eIklAA==",
+      "license": "MIT"
     },
     "node_modules/@types/stack-utils": {
       "version": "2.0.3",
@@ -1275,6 +1302,12 @@
         "sprintf-js": "~1.0.2"
       }
     },
+    "node_modules/asap": {
+      "version": "2.0.6",
+      "resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
+      "integrity": "sha512-BSHWgDSAiKs50o2Re8ppvp3seVHXSRM44cdSsT9FfNEUUZLOGWVCsiWaRPWM1Znn+mqZ1OfVZ3z3DWEzSp7hRA==",
+      "license": "MIT"
+    },
     "node_modules/assertion-error": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/assertion-error/-/assertion-error-1.1.0.tgz",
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -30,14 +30,14 @@
...
-    "@hirosystems/clarinet-sdk": "2.12.0",
+    "@hirosystems/clarinet-sdk": "2.13.0",
     "@stacks/transactions": "^6.16.1",
     "ansicolor": "^2.0.3",
     "fast-check": "^3.20.0",
     "yaml": "^2.6.1"
   },
   "devDependencies": {
-    "@hirosystems/clarinet-sdk-wasm": "2.12.0",
+    "@hirosystems/clarinet-sdk-wasm": "2.13.0",
```
moodmosaic, line 40: Let's wait until 2.14.0 is out in order to not break Node.js 18 and 20 users?

BowTiedRadone: Yes. I think about this, too. We can close this PR, and I will open an issue to add support for remote data 🚀

moodmosaic: Yes, agreed.

BowTiedRadone:
Closing this, as per #116.

BowTiedRadone:
PR reopened. Thanks to @hugocaillard for quickly releasing [a new Clarinet version](https://github.com/hirosystems/clarinet/releases/tag/v2.14.0) with a fix for Node < 22. At this point, the previous Rendezvous functionality should be preserved. **Remote simnet data** is left to be tested before taking this PR out of draft.

citizen.ts
```diff
@@ -54,7 +54,11 @@ export const issueFirstClassCitizenship = async (
     })
   );
 
-  await simnet.initEmptySession();
+  await simnet.initEmptySession({
+    enabled: false,
+    initial_height: 5,
```
moodmosaic, line 59: What 5 means?

citizen.ts
```diff
@@ -54,7 +54,11 @@ export const issueFirstClassCitizenship = async (
...
+    api_url: "",
+  });
```
moodmosaic, line 61: An idea would be to read the session params from rv-config.json (it can be a section there with three key/values) and then:
- If any of them is malformed or missing we throw an error and exit.
- If all of them are valid, we pass them to the empty session.
- If none of them is specified we pass dummy/empty values.

moodmosaic (commented):
🚀

citizen.ts
moodmosaic: So we could rename `path` to `conf` or just keep `path` for now and re-purpose it for `rv-config.json`. Then since it's a JSON, we can deserialize it to a type.

moodmosaic (changes requested):
https://github.com/stacks-network/rendezvous/pull/115/files#r1963796461

citizen.ts
moodmosaic: Yes we can keep it as `path` for now. And if it works we can right after rename to `conf`.

hugocaillard: Theses new settings are in the Clarinet.toml
https://www.hiro.so/blog/test-your-code-with-mainnet-data-in-clarinet-simnet

BowTiedRadone: Thanks for the update and the resources, @hugocaillard. So, to initialize an empty simnet session, can we stick to what's inside the user's `Clarinet.toml`? And override a hardcoded remote data object if there's nothing? Also, what will happen in the scenario used in Rendezvous (redeployment of all contracts by epoch, overwriting the target contract source with the concatenation)?

hugocaillard: You can pass `null` if there's nothing in the clarinet.toml (of if it's not enabled).

> Also, what will happen in the scenario used in Rendezvous (empty session initialization, redeployment of all contracts by epoch, overwriting the target contract source with the concatenation)?

It should just work as usual. But I do encourage you to give it a try

BowTiedRadone: Some updates after trying `initEmptySession` with different `remote_data_settings`:
1. For `null`, the Rendezvous example folder won't be successfully initialized. The error is the following (probably because there is an issue resolving the dependency tree for traits):
```
UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "Contract deployment runtime error: ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.rendezvous-token -> use of undeclared trait <sip-010-trait>".
```
2. For previously working `{ enabled: false, api_url: "", initial_height: 5 }`, same error as for `null`.
3. For remote data (see [Clarinet.toml](https://github.com/stacks-network/rendezvous/pull/115/commits/64ded9f47b4debff77755cccb40c3f212e2ea21b#diff-d254d7f78298bf4facfe0b507b0860069c3fd06ea20bf83ce9a98c8a23a79ec0)), it seems to be working properly. The CI fails because of the API rate limit.

traits.tests.ts
```diff
@@ -2045,15 +2045,14 @@ describe("Trait reference processing", () => {
       },
     };
 
-    const expected = [
+    const expected = new Set([
```
moodmosaic, line 2048: This was (I guess) triggered by a change in the latest clarinet-sdk?

invariant.tests.ts
```diff
@@ -8,7 +8,8 @@ import {
 import { join } from "path";
 import { issueFirstClassCitizenship } from "./citizen";
 import { Cl } from "@stacks/transactions";
-import { getManifestFileName } from "./app";
+import { getManifestFileName, parseRemoteDataSettings } from "./app";
+import { EventEmitter } from "stream";
```
moodmosaic, line 12:
```suggestion
import { EventEmitter } from "events";
```
??

app.ts
```diff
@@ -163,10 +210,13 @@ export async function main() {
...
   }
 
+  const remoteDataSettings = parseRemoteDataSettings(manifestPath, radio);
+
   const simnet = await issueFirstClassCitizenship(
     manifestDir,
     manifestPath,
-    sutContractName
+    sutContractName,
+    remoteDataSettings
```
moodmosaic, line 219: If you move `remoteDataSettings` one position up, which also makes more sense as it's related to the manifest, you'll get a cleaner git diff here.

app.ts
```diff
@@ -10,14 +11,26 @@ import {
...
+ * The object used to initialize an empty simnet session with, when no remote
+ * data is enabled in the `Clarinet.toml` file.
+ */
+export const noRemoteData = {
+  enabled: false,
+  api_url: "",
+  initial_height: 1,
+};
+
+export const invalidRemoteDataErrorMessage = `Remote data settings not properly setup in Clarinet.toml! To use remote data, please provide the "api_url", "initial_height", and "enabled' fields under the "repl.remote_data" section in the Clarinet.toml file.`;
```
moodmosaic, line 32:
```suggestion
export const invalidRemoteDataErrorMessage = `Remote data settings not properly setup in Clarinet.toml! To use remote data, please provide the "api_url", "initial_height", and "enabled" fields under the "repl.remote_data" section in your Clarinet.toml file.`;
```

citizen.ts
hugocaillard: `UnhandledPromiseRejection: "Contract deployment runtime error: ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.rendezvous-token -> use of undeclared trait <sip-010-trait>".`

Does the project pass `clarinet check`?

3. -> rate limiting: you can set an hiro api key in the env var `HIRO_API_KEY`, clarinet will use it and it will increase your rate limit. 
With the nature of Rendezvous itself, it might be very data consuming so even the increased right limit might not be enough, you might have to wait for future iterations of Clarinet tx simulation that will come with some optimizations, especially on data fetching

app.ts
```diff
@@ -41,6 +54,40 @@ export const getManifestFileName = (
   return "Clarinet.toml";
 };
 
+export const parseRemoteDataSettings = (
+  manifestPath: string,
+  radio: EventEmitter
+) => {
```
moodmosaic, line 60: What is the return type of this function? I think this should be renamed to `tryParseRemoteDataSettings` (more on `tryParseXyz` [here](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/exceptions-and-performance#try-parse-pattern)).

app.ts
```diff
@@ -41,6 +54,40 @@ export const getManifestFileName = (
...
+  const clarinetToml = toml.parse(readFileSync(resolve(manifestPath), "utf-8"));
+  const remoteDataUserSettings = clarinetToml.repl
```
moodmosaic, line 62: Wondering it something like null-coalescing works here:

```ts
const remoteDataUserSettings = clarinetToml.repl?.["remote_data"] ?? undefined;
```

app.ts
```diff
@@ -41,6 +54,40 @@ export const getManifestFileName = (
...
+    ? clarinetToml.repl["remote_data"] || undefined
+    : undefined;
+
+  if (
+    remoteDataUserSettings !== undefined &&
+    (!remoteDataUserSettings["initial_height"] ||
+      !remoteDataUserSettings["api_url"] ||
+      !remoteDataUserSettings["enabled"])
+  ) {
```
moodmosaic, line 71: Same here, perhaps (also sorted fields):

```ts
if (
  remoteDataUserSettings?.["api_url"] == null ||
  remoteDataUserSettings?.["enabled"] == null ||
  remoteDataUserSettings?.["initial_height"] == null
) {
  throw new Error(invalidRemoteDataErrorMessage);
}
```

app.ts
```diff
@@ -41,6 +54,40 @@ export const getManifestFileName = (
...
+    throw new Error(invalidRemoteDataErrorMessage);
+  }
+
+  if (remoteDataUserSettings) {
+    radio.emit(
+      "logMessage",
+      yellow(
+        "\nUsing mainnet data. Setting the fuzzing environment will take a while..."
```
moodmosaic, line 79:
```suggestion
        "\nUsing mainnet data. Setting up the environment can take a few minutes..."
```

moodmosaic: Just a thought. What do you think?

citizen.ts
BowTiedRadone: Thanks @hugocaillard! I pushed a potential way of handling Rendezvous + Remote Data in 054f022. I would highly appreciate it if you'd take a look 🙏

app.ts
```diff
@@ -41,6 +54,40 @@ export const getManifestFileName = (
...
+      )
+    );
+  }
+
+  const remoteDataSettings = remoteDataUserSettings
+    ? remoteDataUserSettings
+    : noRemoteData;
```
moodmosaic, line 86: What if we renamed `remoteDataUserSettings` to `remoteData` and then this becomes as

```ts
const remoteDataSettings = remoteData ?? noRemoteData;
```

app.tests.ts
```diff
@@ -1,8 +1,43 @@
...
 import fs from "fs";
+import EventEmitter from "events";
+
+const clarinetTomlRemoteData = {
+  correctSettings: {
+    toml: `
+[repl.remote_data]
+enabled = true
+api_url = 'https://api.hiro.so'
+initial_height = 150000
```
moodmosaic, line 20: I'd sort these, if it doesn't break anything.

app.tests.ts
```diff
@@ -1,8 +1,43 @@
...
+[repl.remote_data]
+enabled = true
+api_url = 'https://api.hiro.so'
+`,
+    expectedRemoteDataSettings: noRemoteData,
+  },
+  noRemoteDataSettings: {
+    toml: ``,
+    expectedRemoteDataSettings: noRemoteData,
+  },
```
moodmosaic, line 39: Nice parameterized test 👍

app.tests.ts
```diff
@@ -422,3 +457,61 @@ describe("Custom manifest detection", () => {
     jest.restoreAllMocks();
   });
 });
+
+describe("Remote data settings parsing", () => {
+  it.each([
+    [
+      "correctly overrides the no remote data settings when the remote data settings are not provided",
+      clarinetTomlRemoteData.noRemoteDataSettings.toml,
+      clarinetTomlRemoteData.noRemoteDataSettings.expectedRemoteDataSettings,
```
moodmosaic, line 466:
```suggestion
      clarinetTomlRemoteData.noRemoteDataSettings.expectedValue,
```

(Maybe?)

app.tests.ts
```diff
@@ -422,3 +457,61 @@ describe("Custom manifest detection", () => {
...
+        initial_height: number;
+      }
+    ) => {
+      // Setup
+      jest
+        .spyOn(fs, "readFileSync")
+        .mockImplementation((path: fs.PathOrFileDescriptor) => tomlContent);
+
+      // Exercise
+      const actual = parseRemoteDataSettings("any path", new EventEmitter());
```
moodmosaic, line 490: We could avoid the `"any path"` if we make the stub strict, e.g.:
```ts
const anyPath = `mock/path/to/${Date.now()}.toml`;
jest
  .spyOn(fs, "readFileSync")
  .mockImplementation((path) => {
    expect(path).toBe(anyPath); // Strict stub.
    return tomlContent;
  });
```

app.tests.ts
```diff
@@ -422,3 +457,61 @@ describe("Custom manifest detection", () => {
...
+  it("throws an error when the remote data settings are not properly set up", () => {
+    // Setup
+    const tomlContent = clarinetTomlRemoteData.noInitialHeightSettings.toml;
+
+    jest
+      .spyOn(fs, "readFileSync")
+      .mockImplementation((path: fs.PathOrFileDescriptor) => tomlContent);
+
+    // Exercise
+    const act = () => parseRemoteDataSettings("any path", new EventEmitter());
```
moodmosaic, line 509: Same here for `"any path"`.

app.ts
moodmosaic: We should add the return type.

traits.tests.ts
BowTiedRadone: Yes. Didn't get the test failure before.

citizen.ts
hugocaillard: Left a few comments there 👍

app.ts
BowTiedRadone: Yes, I agree. The only caveat is that it cannot take a few minutes. Up to a minute should be more realistic. Resolved in 16619a7.

app.ts
BowTiedRadone: Resolved in 4ede957.

app.tests.ts
BowTiedRadone: Resolved in 445f428.

app.tests.ts
BowTiedRadone: Hmmmm. It's actually not a value but an object. Maybe expected?

app.tests.ts
BowTiedRadone: Resolved in 107ce27.

app.tests.ts
BowTiedRadone: Resolved in 107ce27.

app.tests.ts
moodmosaic: Yes 👍👌

app.ts
```diff
@@ -41,6 +55,35 @@ export const getManifestFileName = (
...
+      !remoteDataUserSettings?.["initial_height"])
+  ) {
+    throw new Error(invalidRemoteDataErrorMessage);
+  }
+
+  if (remoteDataUserSettings) {
+    radio.emit(
+      "logMessage",
+      yellow(
+        "\nUsing mainnet data. Setting up the environment can take up to a minute..."
```
moodmosaic, line 79:
```suggestion
        "\nUsing remote data. Setting up the environment can take up to a minute..."
```

Is it always mainnet? Can it be also some prelaunch, tester or other live network? Perhaps _remote_ is a better name?

app.tests.ts
moodmosaic: It’s up to you whether you want to keep the name as-is or rename it. Either way it’s not a blocker.

moodmosaic (approved):
LGTM. We can let @friedger know after we merge, tag, and release. Thanks @hugocaillard for the suggestions.

—

One thing I noticed is that error handling around rate limits could improve—catch, decorate, rethrow, log. (Right now, the user sees only what the underlying clarinet-sdk throws.) This can also be discussed or addressed in a separate PR.

app.ts
BowTiedRadone: Great catch! 🚀

BowTiedRadone:
@hugocaillard, I think @moodmosaic meant to handle the `clarinet-sdk` errors better in Rendezvous. Now, they are reported using the custom reporter as bug findings, but those errors (such as rate limit ones) are not actual findings in the target contract. I believe no action is needed in clarinet.

moodmosaic:
> *[..]* no action is needed *[..]*

Agreed.

PR #126, opened 2025-04-03
[DRAFT] Support sBTC balance for test wallets

### LINKED ISSUE #123: Test wallets sBTC balance is unexpectedly reset
The Rendezvous first-class citizenship issuance process resets the test wallets' sBTC balances, the same as it was happening before for STX–solved by [this PR](https://github.com/stacks-network/rendezvous/pull/111). Now that `clarinet` mints sBTC for test wallets in specific use cases, Rendezvous should keep track of the sBTC balances and ensure the wallets will have the same ones post-redeployment.

BowTiedRadone:
This PR is still a draft. Resolves #123.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.4.1",
       "license": "GPL-3.0-only",
       "dependencies": {
-        "@hirosystems/clarinet-sdk": "2.14.0",
+        "@hirosystems/clarinet-sdk": "2.15.2",
         "@stacks/transactions": "^6.16.1",
         "ansicolor": "^2.0.3",
         "fast-check": "^3.20.0",
@@ -20,7 +20,7 @@
         "rv": "dist/app.js"
       },
       "devDependencies": {
-        "@hirosystems/clarinet-sdk-wasm": "2.14.0",
+        "@hirosystems/clarinet-sdk-wasm": "2.15.2",
         "@types/jest": "^29.5.12",
         "jest": "^29.7.0",
         "ts-jest": "^29.2.5",
@@ -893,12 +893,12 @@
       }
     },
     "node_modules/@hirosystems/clarinet-sdk": {
-      "version": "2.14.0",
-      "resolved": "https://registry.npmjs.org/@hirosystems/clarinet-sdk/-/clarinet-sdk-2.14.0.tgz",
-      "integrity": "sha512-DNIGDIC1kVh0mjBPeIpFdDh+I3+k1nwsED5CCZCnj5d1W+vTWSiREUZLhouq85NahH8j+BPzPZyTEGYbGpocjg==",
+      "version": "2.15.2",
+      "resolved": "https://registry.npmjs.org/@hirosystems/clarinet-sdk/-/clarinet-sdk-2.15.2.tgz",
+      "integrity": "sha512-SPWKYSWI+mvFlZAjzd+cEzAKkHUi30zljA7pxcNupW8z9OszEfYRK13XrfMiZID0cwD5FhH/EyTm/49a8wYRSw==",
       "license": "GPL-3.0",
       "dependencies": {
-        "@hirosystems/clarinet-sdk-wasm": "2.14.0",
+        "@hirosystems/clarinet-sdk-wasm": "2.15.1",
         "@stacks/transactions": "^6.13.0",
         "kolorist": "^1.8.0",
         "prompts": "^2.4.2",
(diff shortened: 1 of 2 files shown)


citizen.ts
```diff
@@ -344,3 +391,125 @@ export function scheduleRendezvous(
...
+        uintCV(amountSats),
+        // (recipient principal)
+        principalCV(recipient),
+        // (burn-hash (buff 32))
+        burnHash.value,
+        // (burn-height uint)
+        uintCV(blockHeight),
+        // (sweep-txid (buff 32))
+        bufferCV(hexToBytes(sweepTxId)),
+      ],
```
hugocaillard, line 478: We tend to prefer Cl helpers; `Cl.uint`, `Cl.buffer` etc (can also avoid the hexToBytes`.
But that's just an alternative, whatever you prefer is fine

citizen.ts
```diff
@@ -50,17 +62,38 @@ export const issueFirstClassCitizenship = async (
...
-  const balancesMap = new Map(
+  const stxBalancesMap = new Map(
     Array.from(simnetAddresses, (address) => {
       const balanceHex = simnet.runSnippet(`(stx-get-balance '${address})`);
       return [address, cvToValue(hexToCV(balanceHex))];
     })
   );
 
+  const sbtcBalancesMap = new Map(
+    Array.from(simnetAddresses, (address) => {
```
hugocaillard, line 73: TIL I learn that you can pass a map function to `Array.from`, but instead, I think it as good to just do `simnetAddress.map(address => { ... })`

citizen.ts
```diff
@@ -344,3 +391,125 @@ export function scheduleRendezvous(
...
+  simnet: Simnet,
+  sbtcBalancesMap: Map<string, number>
+) => {
+  // To deposit sBTC, one needs a txId and a sweep txId. A deposit transaction
+  // must have a unique txId and sweep txId. A set will be used to track the
+  // already used ids. This set will be used in `getUniqueRandomHex`.
+  const usedTxIds: Set<string> = new Set();
+
+  // For each address present in the balances map, restore the balance.
+  Array.from(sbtcBalancesMap.entries())
```
hugocaillard, line 413: You don't need `array.from` here do you?
entries() already returns an array

citizen.ts
```diff
@@ -344,3 +391,125 @@ export function scheduleRendezvous(
...
+  amountSats: number,
+  recipient: string,
+  txId: string,
+  sweepTxId: string
+) => {
+  const blockHeight = simnet.blockHeight;
+
+  // Calling `get-burn-header` only works for past block heights. We mine one
+  // empty Bitcoin block along with a Stacks one to make sure this happens.
+  simnet.mineEmptyBurnBlocks(1);
```
hugocaillard, line 443: `mintSBtc()` is called for every account that needs to be founded, so if there are 10 accounts you'll mine 10 burn blocks, if there are 2 accounts you'll mine 2 burn blocks.
That might be weird.
I think you could simple pass `Cl.uint(simnet.burnBlockHeight - 1)`

citizen.ts
```diff
@@ -344,3 +391,125 @@ export function scheduleRendezvous(
...
+  // Keep trying to generate random hex strings until the generated one is not
+  // present in the `usedTxIds` set.
+  do {
+    // Generate a 32-byte (64 character) random hex string.
+    const bytes = new Uint8Array(32);
+    crypto.getRandomValues(bytes);
+    hex = Array.from(bytes)
+      .map((byte) => byte.toString(16).padStart(2, "0"))
+      .join("");
+  } while (usedTxIds.has(hex));
```
hugocaillard, line 510: Is this necessary? What are the odds of having a collision?

hugocaillard (commented):
Looking good, all comment are only suggestions

citizen.ts
moodmosaic: If that’s isomorphic (produces same output) we should do it in all the other places too (in a separate commit/PR). 👍

citizen.ts
BowTiedRadone: If the same `txid` is used twice the deposit [fails](https://github.com/stacks-sbtc/sbtc/blob/f39002009268554198f49e99959c760ef1a18104/contracts/contracts/sbtc-deposit.clar#L56) with `(err 301)`. Initially, I tried using the same one for all the deposits, but only the first one ended up being funded.

hugocaillard: Yes the crypto module will give you a unique thing everytime

moodmosaic: However, it won’t hurt to keep the check

`usedTxIds.has(hex)`

— Or remove the check and use a self-describing function like generateUniqueXyz.

- - -

I assume some (future) contributors reading the code might not be instantly familiar with how that works.

BowTiedRadone: Yes, @moodmosaic. Now it's a mix of self-describing function and loop. So I'll remove the loop.

moodmosaic (approved):
Thanks @hugocaillard for looking into this!

LGTM (too).

Issue #129, opened 2025-04-08
Disable remote data by default if `[repl.remote_data]` is missing data


BowTiedRadone:
If the user's `Clarinet.toml` doesn't have enough info for the simnet remote data feature, Rendezvous will throw an error. That can be inconvenient. Fall back to running with remote data disabled instead.

Thanks to @friedger for flagging this!

PR #131, opened 2025-04-08
Improve remote data settings processing

BowTiedRadone:
This PR updates how remote data settings are handled. If `Clarinet.toml` is missing or has incomplete remote data fields, it logs a warning instead of throwing an error. This makes it easier to run Rendezvous without needing to fix the manifest file when remote data isn't fully set up.

Sample output for incomplete remote data settings:
```
Using manifest path: ../clarity-catamaranswaps/catamaran-sbtc/Clarinet.toml
Target contract: btc-sbtc-swap

Remote data settings existing in Clarinet.toml, but remote data feature will not be used! To use remote data, please make sure the following fields are set:

- enabled = true
- api_url = <stacks-api-url>
- initial_height = <stacks-block-height-to-fork-from>

under the "repl.remote_data" section in the Clarinet.toml file.

Starting invariant testing type for the btc-sbtc-swap contract...

```

Resolves https://github.com/stacks-network/rendezvous/issues/129.


app.tests.ts
```diff
@@ -38,6 +38,15 @@ enabled = true
...
   },
+  enabledFalseSettings: {
+    toml: `
+[repl.remote_data]
+api_url = 'https://api.hiro.so'
+enabled = false
+initial_height = 595012
+`,
+    expected: noRemoteData,
+  },
```
BowTiedRadone, line 49: @friedger Here you can see your specific settings. Used them in a test to make sure it'll work fine for your [catamaranswaps](https://github.com/friedger/clarity-catamaranswaps) sBTC use case.

moodmosaic (approved):
Good catch! LGTM 👍

PR #136, opened 2025-04-14
Show function call results in run output lines

### LINKED ISSUE #132: Show actual function results in `Outputs` section on failure
Right now, when a test fails, the counterexample includes an `Outputs` section. But this section shows the return **type** of each function, not the actual result of the call. It would be helpful to see the real Clarity snippet outputs—like `(ok true)`—to better understand what went wrong.

Thanks to @friedger for flagging this!

BowTiedRadone:
This PR improves Rendezvous logging by adding the actual Clarity result to each output line for public functions, invariants, and property tests. Now, users can see the exact return value from each call directly in the logs.

For example, logs now show `(ok true)` for passing property tests, `(ok false)` for discarded, and the specific Clarity snippets for the failing ones. Public function call and invariant output lines also include their actual results during the invariant testing runs.

Resolves #132.

### DIFF
--- a/invariant.ts
+++ b/invariant.ts
@@ -6,7 +6,7 @@ import {
   getFunctionsListForContract,
 } from "./shared";
 import { LocalContext } from "./invariant.types";
-import { Cl, cvToJSON } from "@stacks/transactions";
+import { Cl, cvToJSON, cvToString } from "@stacks/transactions";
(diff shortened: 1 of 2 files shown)


moodmosaic (approved):
LGTM 👍 💯

Issue #127, opened 2025-04-07
Periodic housekeeping


BowTiedRadone:
Thanks to the latest efforts things are moving fast, hopefully for the good of Stacks ecosystem's security. However, when new features are added periodic housekeeping is required.

The housekeeping should target items like:

- [x] [this PR comment](https://github.com/stacks-network/rendezvous/pull/126#discussion_r2029271952)
- [x] book refinements [here](https://stacks-network.github.io/rendezvous/chapter_3.html#when-typescript-testing-excels)
- [x] adding book link to `README`
- [x] go through https://github.com/stacks-network/rendezvous/security and related PRs

BowTiedRadone:
Housekeeping completed.

PR #138, opened 2025-04-15
Book housekeeping

BowTiedRadone:
This PR addresses a misleading detail found in the book, flagged by @hugocaillard. Related to #127.


BowTiedRadone:
@moodmosaic Before merging we should check if the docs publishing action still works. If not, we should tackle it in this PR, too.

moodmosaic (approved):
👍 📖 📕

PR #164, opened 2025-07-28
Add fail-fast option

### LINKED ISSUE #161: Add `fail-fast`/`no-shrink` option
Sometimes shrinking hides the first failure of a test run and reports only the last one. To enable users to check it, we could add a `fail-fast`/`no-shrink` option.

BowTiedRadone:
This PR adds a fail-fast flag (`--bail`) so users can stop fuzzing as soon as something breaks instead of going through the rest of the run and shrinking.

Resolves #161.

### DIFF
--- a/app.tests.ts
+++ b/app.tests.ts
@@ -150,6 +150,16 @@ describe("Command-line arguments handling", () => {
       ],
     ],
     [
+      ["manifest path", "contract name", "seed", "bail"],
+      ["node", "app.js", "example", "counter", "--bail"],
+      [
+        red(
+          `\nInvalid type provided. Please provide the type of test to be executed. Possible values: test, invariant.`
+        ),
+        helpMessage,
+      ],
+    ],
+    [
       ["manifest path", "contract name", "seed"],
       ["node", "app.js", "example", "counter", "--seed=123"],
       [
@@ -226,6 +236,16 @@ describe("Command-line arguments handling", () => {
       ],
     ],
     [
+      ["manifest path", "contract name", "type=invariant", "bail"],
+      ["node", "app.js", "example", "counter", "invariant", "--bail"],
+      [
+        `Using manifest path: example/Clarinet.toml`,
+        `Target contract: counter`,
+        `Bailing on first failure.`,
+        `\nStarting invariant testing type for the counter contract...\n`,
+      ],
+    ],
+    [
       ["manifest path", "contract name", "type=invariant", "dialers file path"],
       [
         "node",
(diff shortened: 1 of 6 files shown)


app.tests.ts
```diff
@@ -52,23 +52,6 @@ initial_height = 595012
 describe("Command-line arguments handling", () => {
   const initialArgv = process.argv;
 
-  const helpMessage = `
```
moodmosaic, line 55: At first glance, removing the test message from here (and making the other one public) makes the tests seem less trustworthy. If the app's help message changes, will the tests break?

app.ts
```diff
@@ -106,11 +106,15 @@ const helpMessage = `
     --seed - The seed to use for the replay functionality.
     --path - The path to use for the replay functionality.
     --runs - The runs to use for iterating over the tests. Default: 100.
+    --bail - Stop after the first failure.
```
moodmosaic, line 109: 🥇

app.ts
```diff
@@ -106,11 +106,15 @@ const helpMessage = `
...
     --dial – The path to a JavaScript file containing custom pre- and post-execution functions (dialers).
     --help - Show the help message.
   `;
 
-const parseOptionalArgument = (argName: string) => {
+const parseBooleanOption = (argName: string): boolean =>
```
moodmosaic, line 114: The rule of three is starting to kick in. Eventually -- though it's not a high priority right now -- we should consider using some battle-hardened, structured command-line argument libraries for JavaScript.

Perhaps we can open an issue for this.

property.ts
```diff
@@ -29,6 +29,7 @@ import {
  * @param seed The seed for reproducible property-based tests.
  * @param path The path for reproducible property-based tests.
  * @param runs The number of test runs.
+ * @param endOnFailure Whether to stop fuzzing after the first failure.
```
moodmosaic, line 32: Is this precise? I was thinking of something along the lines of 'do not attempt to shrink' after the first failure, since shrinking kicks in after a failure occurs.

```suggestion
 * @param endOnFailure Stop execution after the first failure and prevent further shrinking.
```

property.ts
```diff
@@ -40,6 +41,7 @@ export const checkProperties = (
   seed: number | undefined,
   path: string | undefined,
   runs: number | undefined,
+  endOnFailure: boolean,
```
moodmosaic, line 44: Shouldn't this be `bail` to match _our_ DSL? Internally, it can be mapped to whatever fast-check expects. This prevents us from exposing implementation details, avoiding a [leaky abstraction](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/).

```suggestion
  bail: boolean,
```

BowTiedRadone:
> > After we agree on this PR's content we have to update the docs.
> 
> Good point! Perhaps we can do it here in this PR?

Sure!

app.ts
BowTiedRadone: Yes. I will open an issue for this.

BowTiedRadone: Done. Opened #165.

docs/chapter_6.md
```diff
@@ -172,7 +172,17 @@ In this case, the seed is `426141810`. You can use it to rerun the exact same te
...
-**3. Using Dialers**
+**3. Stop After First Failure**
+
+By default, Rendezvous will start the shrinking process after finding a failure. To stop immediately when the first failure is detected, use the `--bail` option:
+
+```bash
+rv root contract test --bail
+```
+
+This is useful when you want to examine the first failure without waiting for the complete test run and shrinking process to finish.
```
moodmosaic, line 183: 💯

moodmosaic (approved):

PR #169, opened 2025-08-18
Document random trait reference selection

### LINKED ISSUE #158: Add random trait reference selection section to the docs
It's unclear how Rendezvous selects trait references from the Clarinet project during random argument generation. Add a section explaining this topic to the docs.

BowTiedRadone:
This PR documents how Rendezvous discovers and selects trait implementations when SUT/test/invariant functions accept trait reference parameters.

Resolves #158.

### DIFF
--- a/docs/chapter_6.md
+++ b/docs/chapter_6.md
@@ -527,7 +527,7 @@ This ensures that the test double is only used when testing `sbtc-token`, keepin
 
 ## Trait Reference Parameters
 
-Rendezvous automatically generates arguments for function calls using [fast-check](https://github.com/dubzzz/fast-check). It handles most Clarity types without any setup from you. However, **trait references** require special handling since Rendezvous cannot generate them automatically.
+Rendezvous automatically generates arguments for function calls. It handles most Clarity types without any setup from you. However, **trait references** require special handling since Rendezvous cannot generate them automatically.
 
 ### How Trait Reference Selection Works
 
--- a/docs/chapter_6.md
+++ b/docs/chapter_6.md
@@ -545,8 +545,8 @@ This process allows Rendezvous to create meaningful state transitions and valida
 
 The `example` Clarinet project demonstrates this feature. The [send-tokens](https://github.com/stacks-network/rendezvous/blob/9c02aa7c2571b3795debc657bd433fd9bf7f19eb/example/contracts/send-tokens.clar) contract contains [one public function](https://github.com/stacks-network/rendezvous/blob/9c02aa7c2571b3795debc657bd433fd9bf7f19eb/example/contracts/send-tokens.clar#L3-L7) and [one property-based test](https://github.com/stacks-network/rendezvous/blob/9c02aa7c2571b3795debc657bd433fd9bf7f19eb/example/contracts/send-tokens.tests.clar#L24-L47) that both accept trait references.
 
-To enable testing, the project includes [rendezvous-token](https://github.com/stacks-network/rendezvous/blob/9c02aa7c2571b3795debc657bd433fd9bf7f19eb/example/contracts/rendezvous-token.clar), which implements the required trait. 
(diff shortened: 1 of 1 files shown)


docs/chapter_6.md
```diff
@@ -519,3 +524,29 @@ epoch = 3.0
 - If not, it **falls back** to the standard `Clarinet.toml`.
 
 This ensures that the test double is only used when testing `sbtc-token`, keeping tests realistic while allowing necessary state transitions.
+
+## Trait Reference Parameters
+
+Rendezvous automatically generates arguments for function calls using [fast-check](https://github.com/dubzzz/fast-check). It handles most Clarity types without any setup from you. However, **trait references** require special handling since Rendezvous cannot generate them automatically.
```
moodmosaic, line 530:
```suggestion
Rendezvous automatically generates arguments for function calls. It handles most Clarity types without any setup from you. However, **trait references** require special handling since Rendezvous cannot generate them automatically.
```

moodmosaic: fast-check is an implementation detail -- in the future we might switch to something else. Unlikely, but better not hardcode an implementation detail in the docs?

moodmosaic: Also, in _"It handles most Clarity types without any setup from you."_ once we add support for custom generators we'd have to remember to go back and update this. Perhaps we can remove this.

docs/chapter_6.md
```diff
@@ -519,3 +524,29 @@ epoch = 3.0
...
+### How Trait Reference Selection Works
+
+When your functions accept trait reference parameters, you must include at least one trait implementation in your Clarinet project. This can be either a project contract or a requirement.
+
+Here's how Rendezvous handles trait references:
+
+1. **Project Scanning** – Before testing begins, Rendezvous scans your project for functions that use trait references.
+2. **Implementation Discovery** – It searches the contract AST for matching trait implementations and adds them to a selection pool.
+3. **Random Selection** – During test execution, Rendezvous randomly picks an implementation from the pool and uses it as a function argument.
+
```
moodmosaic, line 541: This reminds me what a huge amount of work has been put in! ❤️

moodmosaic: Perhaps in `3.` we can mention that by replaying via a seed, the same trait implementation will be picked up.

docs/chapter_6.md
```diff
@@ -519,3 +524,29 @@ epoch = 3.0
...
+### Example
+
+The `example` Clarinet project demonstrates this feature. The [send-tokens](https://github.com/stacks-network/rendezvous/blob/9c02aa7c2571b3795debc657bd433fd9bf7f19eb/example/contracts/send-tokens.clar) contract contains [one public function](https://github.com/stacks-network/rendezvous/blob/9c02aa7c2571b3795debc657bd433fd9bf7f19eb/example/contracts/send-tokens.clar#L3-L7) and [one property-based test](https://github.com/stacks-network/rendezvous/blob/9c02aa7c2571b3795debc657bd433fd9bf7f19eb/example/contracts/send-tokens.tests.clar#L24-L47) that both accept trait references.
+
+To enable testing, the project includes [rendezvous-token](https://github.com/stacks-network/rendezvous/blob/9c02aa7c2571b3795debc657bd433fd9bf7f19eb/example/contracts/rendezvous-token.clar), which implements the required trait. 
+
+### Adding More Implementations
+
+You can include multiple trait implementations in your project. The more implementations you add, the more randomness Rendezvous can introduce during testing. This helps ensure your contract works correctly with different trait implementation contracts.
```
moodmosaic, line 552: However you'd have to make sure to increase the number of runs -- or somehow add how the number of runs correlates to this.

docs/chapter_6.md
BowTiedRadone: I think the second can stay, as the default generators still won't require user interaction. I see an eventual generators customization as something extra the user can choose to do.

docs/chapter_6.md
BowTiedRadone: Not 100% sure about this. We already have a seed section–wouldn't repeating how seed works be redundant here?

docs/chapter_6.md
moodmosaic: We should explain why that happens. And also add that as it’s valuable info. 👍

PR #176, opened 2025-09-17
Switch npm publish workflow to use OIDC trusted publishing

BowTiedRadone:
This PR removes the use of the `NPM_TOKEN` secret and enables npm trusted publishing with provenance using GitHub Actions OIDC.


wileyj (approved):
per the docs, lgtm

PR #178, opened 2025-09-24
Simplify remote data parsing

BowTiedRadone:
This PR fixes an inconsistency when users omit `repl.remote_data.api_url` from their Clarinet configuration, while still specifying the other fields.

Clarinet defaults to the mainnet API URL when `api_url` is unset. Rendezvous treated this as if remote data was disabled and initialized an empty session without using remote data. This caused deployment failures because Rendezvous couldn't find requirements that Clarinet expected to be on mainnet

This updates Rendezvous so it matches Clarinet's behavior.

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -57,9 +57,8 @@ export const tryParseRemoteDataSettings = (
 ): RemoteDataSettings => {
   const clarinetToml = toml.parse(readFileSync(resolve(manifestPath), "utf-8"));
   const remoteDataUserSettings = clarinetToml.repl?.remote_data ?? undefined;
-  const remoteDataDisabled = !remoteDataUserSettings?.enabled;
 
-  if (remoteDataUserSettings && !remoteDataDisabled) {
+  if (remoteDataUserSettings && remoteDataUserSettings?.enabled === true) {
     radio.emit(
       "logMessage",
       yellow(


app.ts
```diff
@@ -69,8 +68,11 @@ export const tryParseRemoteDataSettings = (
   const remoteDataUserSettings =
     clarinetToml.repl?.["remote_data"] ?? undefined;
 
+  // Default to Hiro API when no URL is provided (matches Clarinet behavior).
+  // This ensures correct deployment plan processing and remote data decisions.
+  const apiUrl = remoteDataUserSettings?.["api_url"] || "https://api.hiro.so";
```
moodmosaic, line 73: Any concerns about rate limits? Otherwise LGTM.

hugoclrd (changes requested):
I don't think that rendezvous needs to set the default URL.
Does rendezvous itself need to make API calls? If not, it should probably let Clarinet do its things

app.ts
hugoclrd: Clarinet MXS supports API keys. It'll take the one set in the env var `HIRO_API_KEY` if any

BowTiedRadone:
> @BowTiedRadone Yes, this is the default API URL. Why does rendezvous needs to know this URL? Does it fetch data itself?

@hugoclrd Rendezvous needs a remote data object to start an empty simnet session. [Here](https://github.com/friedger/clarity-ccip-026/blob/300c43fa8090c75cf03b9c0fed4522fd36eaf811/Clarinet.toml#L38-L43), the remote data didn’t specify an `api_url`. Since Rendezvous couldn’t find one, it used the default for missing remote data, while simnet defaulted to the mainnet API URL. This led to an inconsistency (Rendezvous could not deploy one of the contracts because the deployment plan expectedly did not include the requirements).

hugoclrd:
Ok, I see.
So you can call `initEmptySession()` like so
```ts
await simnet.initEmptySession({ enabled: true })
```

So that you'll let Clarinet handle the default for the API URL, and more importantly, for the block height (which is dynamic - clarinet fetches the latest)

moodmosaic (commented):
I lean towards https://github.com/stacks-network/rendezvous/pull/178/files#r2377927257.

app.ts
```diff
@@ -85,7 +68,10 @@ export const tryParseRemoteDataSettings = (
     );
   }
 
-  if (!remoteDataUserSettings || invalidRemoteDataSetup) {
+  // If no remote data settings are provided, we still need to return an object
+  // with the `enabled` property set to `false`. That is what simnet expects
+  // at least in order to initialize an empty simnet session.
+  if (!remoteDataUserSettings) {
```
moodmosaic, line 74: What if a remote data is provided but not in valid form?

It looks like this removes the validation for `api_url` and `initial_height`, which allows an incomplete `remote_data` configuration to be passed downstream(?)

moodmosaic (commented):
https://github.com/stacks-network/rendezvous/pull/178/files#r2378627830

app.ts
BowTiedRadone: In the latest version of `clarinet-sdk`, `initial_height` and `api_url` are optional. They will default to whatever simnet defaults as usual. Check [this](https://github.com/stacks-network/rendezvous/pull/178#issuecomment-3330006320) for more info.

hugoclrd:
Love the latest changes 👍 

As a side note, that rendezvous is one of the few maintained project that build a tool on top of the SDK (since 99% of the SDK users use it to write unit tests).

So if you ever think that it could expose more data, or that the API should be different, or that the settings management should be better, etc, please start by opening an issue on Clarinet so that we can try to fix upstream 🙏 We can release fixes pretty quickly.

Patching it in rendezvous should always be the last resort

Issue #172, opened 2025-09-10
The CI is intermittently failing with "Unable to parse metadata file"


BowTiedRadone:
The CI is intermittently failing in `heatstroke.tests.ts` with the following error:
```
FAIL ./heatstroke.tests.ts (11.653 s)
  ● Custom reporter logging › handles cases with missing path on failure for invariant testing type

    thrown: "Unable to parse metadata file: EOF while parsing a value at line 1 column 0"

       8 |
       9 | describe("Custom reporter logging", () => {
    > 10 |   it("handles cases with missing path on failure for invariant testing type", async () => {
         |   ^
      11 |     const manifestPath = resolve(__dirname, "./example/Clarinet.toml");
      12 |     const simnet = await initSimnet(manifestPath);
      13 |

      at heatstroke.tests.ts:10:3
      at Object.<anonymous> (heatstroke.tests.ts:9:1)
```
Example of a failed job:
https://github.com/stacks-network/rendezvous/actions/runs/17616671377/job/50051518595

PR #182, opened 2025-09-25
Remove test race conditions using isolated Clarinet environments

BowTiedRadone:
Tests that initialize simnet concurrently were causing "EOF while parsing metadata" errors due to shared file access. Each test now gets its own isolated copy of the example project, eliminating race conditions. Addresses #172.


moodmosaic (commented):
This looks good. The only reason I can see for the draft status is the incomplete teardown in some tests.

BowTiedRadone:
> This looks good. The only reason I can see for the draft status is the incomplete teardown in some tests.

That was not actually the case. The teardown was already in place on master, so it doesn’t show up in the diff. The real issue was CI failing even after the changes, which should now be fixed. That said, GitHub reported some workflow issues causing macOS jobs to stay pending: https://www.githubstatus.com/incidents/071h21gptcp0

Once CI is green, this PR can come out of draft.

wileyj (approved):
lgtm

PR #184, opened 2025-09-26
Automate example project updates after npm release

BowTiedRadone:
This PR updates the npm publish workflow to automatically update the example project with the latest published version. This keeps the example in sync without manual intervention. It also updates `example` project's `package-lock.json`.

### DIFF
--- a/.github/workflows/npm-publish-release.yml
+++ b/.github/workflows/npm-publish-release.yml
@@ -23,3 +23,79 @@ jobs:
       - run: npm test
       - run: npm run build
       - run: npm publish
+
+  update-example-post-release:
+    needs: publish
+    runs-on: ubuntu-latest
+    env:
+      VERSION: ${{ github.ref_name }}
+      BRANCH_NAME: post-release/update-example
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+          registry-url: "https://registry.npmjs.org"
+
+      - name: Wait for npm propagation
+        run: sleep 30
+
+      - name: Verify version availability on npm with retries
+        run: |
+          echo "Checking if @stacks/rendezvous@$VERSION is available on npm..."
+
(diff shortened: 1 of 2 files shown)


.github/workflows/npm-publish-release.yml
```diff
@@ -24,3 +20,19 @@ jobs:
...
+
+          cd example
+          npm install @stacks/rendezvous@latest
+          cd ..
+
+          # Commit the updated package-lock.json.
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git add example/package-lock.json
+          git commit -m "chore: update example to use latest published version" || exit 0
```
moodmosaic, line 37:
```suggestion
          git commit -m "Update example to use latest published version" || exit 0
```

moodmosaic:
This is nice to have. 👌

moodmosaic:
Sounds good and at the same time a bit complicated. Can't we have the workflow commit to a release branch instead? And then merge the PR. It can be when we do a release PR for example.

BowTiedRadone:
> Sounds good and at the same time a bit complicated. Can't we have the workflow commit to a release branch instead? And then merge the PR. It can be when we do a release PR for example.

@moodmosaic Yes, that’s what I’m proposing too, but with automated PR creation to remove as many manual steps as possible. For example, once the release workflow detects the deployed release on npm, a post-release workflow can update the example project, push to a post-release branch, and open a PR for review. This way we keep the workflows focused and single-responsible.

moodmosaic:
If there's a way to do this in one shot without relying on the NPM registry/API (which can be flaky at times) it'd be easier. I think, we always have a `release*` branch. Perhaps, a workflow could push to that branch.

wileyj (commented):
the proposed seems fine, but it makes me wonder if it would be a better idea to simply use the local repo for the examples vs updating the npmjs rendezvous version?

wileyj (approved):

Issue #191, opened 2025-10-19
Rendezvous v1.0.0 Roadmap


BowTiedRadone:
Version 1.0.0 consolidates the must-have features, fixes, and improvements needed to make Rendezvous stable and production-ready as a robust fuzzer for Clarity smart contracts. This release focuses on core functionality, reliability, feature completeness, documentation, and overall developer experience.

### Bug Fixes
* [x] #157
* [x] #174
* [x] #219

### UX
* [ ] #257 (BREAKING)
* [x] ~#60~ (postponed)
* [x] #130
* [x] ~#148~ (postponed) 
* [x] #183 
* [x] Quickstart tutorial
* [x] Smooth integration with Clarinet (once https://github.com/stx-labs/clarinet/issues/2022 is implemented)

### Maintainability
* [x] #165
* [x] #180
* [x] #181
* [x] #194

### Clarity Property-Based Testing
* [x] #168

## Notes

- Although #7 is currently closed, reopening it is being considered. While #60 and #66 target similar outcomes, a configuration file may offer a cleaner and more flexible approach for Rendezvous users, especially in scenarios where new configuration options may be introduced as part of #148 and #183.

---

This list may evolve as development progresses and priorities are refined.

BowTiedRadone:
Since there was no real ask for #60 and #148, and they are not going to affect backwards compatibility, I decided to postpone them for a post-`1.0.0` future version.

PR #195, opened 2025-10-28
Fix trait detection for defined traits and improve handling

### LINKED ISSUE #157: Handle missing trait implementations gracefully instead of crashing
Rendezvous crashes when it encounters a function with trait reference parameters but no trait implementation exists in the project. It runs tests and eventually selects this function for random argument generation, which causes a crash. We should handle missing trait implementations gracefully instead of crashing during the test run.

Possible solutions:
- Remove problematic functions from selection pool - print a warning and throw an error if no other SUT functions/invariants/tests are found
- Fail fast - throw an error immediately when the missing implementation is detected

The first solution makes onboarding easier for new users or for adding Rendezvous to projects, but it can hide the fact that trait reference functions are excluded from the test run.

BowTiedRadone:
This PR fixes a bug in the trait implementation detection logic that prevented Rendezvous from correctly identifying contracts implementing a trait defined with `(define-trait <trait_name> (<trait_definition>))`. This issue caused runs to fail, partly due to #157 (Rendezvous does not handle missing trait implementations gracefully and crashes mid-run).

Resolves #157 (part of #191).

### DIFF
--- a/invariant.ts
+++ b/invariant.ts
@@ -59,15 +59,16 @@ export const checkInvariants = async (
       failed: new Map<string, number>(),
     },
   };
+  // The Rendezvous identifier is the first one in the list. Only one contract
+  // can be fuzzed at a time.
+  const rendezvousContractId = rendezvousList[0];
+
   // A map where the keys are the Rendezvous identifiers and the values are
   // arrays of their SUT (System Under Test) functions. This map will be used
   // to access the SUT functions for each Rendezvous contract afterwards.
   const rendezvousSutFunctions = filterSutFunctions(rendezvousAllFunctions);
 
-  // The Rendezvous identifier is the first one in the list. Only one contract
-  // can be fuzzed at a time.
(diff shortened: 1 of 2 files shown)


invariant.ts
```diff
@@ -88,49 +89,29 @@ export const checkInvariants = async (
     statistics.invariant!.failed.set(functionInterface.name, 0);
   }
 
-  const traitReferenceSutFunctions = rendezvousSutFunctions
-    .get(rendezvousContractId)!
-    .filter((fn) => isTraitReferenceFunction(fn));
+  const sutFunctions = rendezvousSutFunctions.get(rendezvousContractId)!;
```
hugo-stacks: For my on knowledge, what does `sut` mean?

BowTiedRadone: System Under Test. During invariant testing runs, functions from the original contract (`SUT`) are called randomly to create state transitions.

PR #196, opened 2025-10-30
Document update

mahanGurung:
## Description

This is needed because it is latest stack.js type system. This pull request changes the `9 and 10` value to human-readable representation of Clarity values like in the https://github.com/stacks-network/rendezvous/blob/master/example/sip010.cjs.

```
 // If the memo argument is `none`, there's nothing to validate.
  if (memoGeneratedArgumentCV.type === "none") return;

  // Ensure the memo argument is an option (`some`).
  if (memoGeneratedArgumentCV.type !== "some") {
    throw new Error("The memo argument must be an option type!");
  }
```

## Type of Change

- Documentation update


BowTiedRadone:
@mahanGurung Could you please sign your commit? Rendezvous requires all commits to be signed. Once that’s updated, I’ll merge your PR.

PR #205, opened 2025-11-20
CI housekeeping

BowTiedRadone:
This PR pins external CI actions to specific commit SHAs and updates them to their latest versions. It also refactors the docs workflow to always build but only deploy on the `master` branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Rendezvous `rv` is a Clarity fuzzer designed to cut through your smart contract'
 
 ### Prerequisites
 
-- **Node.js**: Supported versions include 20, 22, and 23. Other versions may work, but they are untested.
+- **Node.js**: Supported versions include 20, 22, and 24. Other versions may work, but they are untested.
 
 ### Inspiration
 
--- a/docs/chapter_5.md
+++ b/docs/chapter_5.md
@@ -32,7 +32,7 @@ This chapter covers how to install Rendezvous and set up your environment for ef
 
 Before installing Rendezvous, ensure you have the following prerequisites:
 
-- **Node.js**: Rendezvous supports Node.js versions 20, 22, and 23. Other versions may work but are untested.
+- **Node.js**: Rendezvous supports Node.js versions 20, 22, and 24. Other versions may work but are untested.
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -15,14 +15,14 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        node: [23, 22, 20]
+        node: [24, 22, 20]
```
moodmosaic: Looks correct according to https://nodejs.org/en/about/previous-releases 👍

.github/workflows/ci.yml
```diff
@@ -73,30 +73,30 @@ jobs:
...
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
 
       - name: Setup mdBook
-        uses: peaceiris/actions-mdbook@v2
+        uses: peaceiris/actions-mdbook@ee69d230fe19748b7abf22df32acaa93833fad08 # v2.0.0
         with:
-          mdbook-version: "0.5.0"
+          mdbook-version: "latest"
```
moodmosaic: Can you use a git commit hash here as above? Otherwise LGTM.

BowTiedRadone:
@moodmosaic Thanks for the review! What's left is to update the supported Node versions in README and docs. Still thinking about fixed `0.5.0` vs `latest` (reproducibility being the pro for fixed one), since a SHA [does not seem to be working](https://github.com/peaceiris/actions-mdbook/blob/c55832435624ab1e08e67c99dbd044c768855937/src/get-url.ts#L13) with this action. Also, I'm challenging the idea itself of continuing using https://github.com/peaceiris/actions-mdbook, since it seems to no longer be maintained (issues and dependabot PRs look concerning).

.github/workflows/ci.yml
wileyj: wouldnt this commit have the same issue that required pinning the version? isee the action commit is from 2024

BowTiedRadone: The issue was fixed in this commit: https://github.com/stacks-network/rendezvous/commit/e5eb984f0efd66f920ad9776029ed519a02737a0 (preceding the version pin one). It consisted of the `multilingual` field being dropped in `0.5.0`. However, for reproducibility I think it makes sense to drop this change and pin `0.5.0`.

Issue #180, opened 2025-09-25
Remove sBTC terminology from public function/API surface


moodmosaic:
before `rendezvous` v1 is released.

Right now, some public-facing/exported functions and tests in `rendezvous` use sBTC-specific naming (e.g. `getSbtcBalancesFromSimnet`). This leaks application-level terminology into what should remain a generic library.

sBTC is more of an app, and tying `rendezvous`’s exported functions to it creates coupling we’ll likely regret—for example, if sBTC moves to a different version (v2, v3, etc.). Instead, references to sBTC should remain internal/implementation details.

This minimizes the surface area tied to a specific app, keeps `rendezvous` a proper generic library, and avoids future migration issues when apps like sBTC evolve. See also: https://github.com/stacks-network/rendezvous/pull/179#discussion_r2377991069.

PR #206, opened 2025-11-27
Run Rendezvous in temporary isolated Clarinet projects

BowTiedRadone:
This PR is meant to simplify the Rendezvous tests first class citizenship logic by passing the deployment plan responsibility down to `clarinet-sdk`. It brings Rendezvous closer to clarinet by working *with* it rather than *around* it, aiming to minimize the internal complexity and improve Rendezvous' overall reliability.

Closes https://github.com/stacks-network/rendezvous/issues/180 as a side effect of the refactoring.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   tests:
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 30
+    timeout-minutes: 10
     strategy:
       fail-fast: false
       matrix:
--- a/jest.config.js
+++ b/jest.config.js
@@ -7,7 +7,7 @@ module.exports = {
   },
   modulePathIgnorePatterns: ["<rootDir>/dist/"],
   setupFilesAfterEnv: ["<rootDir>/jest.setup.js"],
-  testTimeout: 1800000, // 30 minutes
+  testTimeout: 600000, // 10 minutes
   maxWorkers: 1,
   collectCoverage: false,
   collectCoverageFrom: [


hugo-stacks:
Since this will be implemented in clarinet (https://github.com/stx-labs/clarinet/issues/2022), my recommendation would be to pause in rendezvous for now

BowTiedRadone:
@hugo-stacks I decided to try and merge this, as it fixes an existing issue. This also simplifies the codebase, preparing it for when https://github.com/stx-labs/clarinet/issues/2022 is ready.

citizen.ts
```diff
@@ -1,283 +1,138 @@
...
-              path: contract.path,
-              clarity_version: contract["clarity-version"],
-            },
-          };
-        });
-
-      if (contracts.length > 0) {
-        acc[epoch] = (acc[epoch] || []).concat(contracts);
-      }
+  const sessionId = `${Date.now()}`;
```
moodmosaic: Why not use proper unique IDs?
```suggestion
  const sessionId = randomUUID();
```

citizen.ts
```diff
@@ -1,283 +1,138 @@
...
+  const temporaryContractsDir = join(manifestDir, ".rv-contracts", sessionId);
+  mkdirSync(temporaryContractsDir, { recursive: true });
```
moodmosaic: It might be worth considering doing some cleanup here for orphaned files, e.g.:

```js
// Add startup cleanup for orphaned files.
function cleanupOrphanedTempFiles(manifestDir: string) {
  const tempManifestPattern = /^\.Clarinet\.toml\.\w+$/;
  // Clean up old temp files...
}
```

citizen.tests.ts
```diff
@@ -1,21 +1,18 @@
...
   getTestContractSource,
-  groupContractsByEpochFromDeploymentPlan,
   issueFirstClassCitizenship,
   scheduleRendezvous,
 } from "./citizen";
 import { initSimnet } from "@stacks/clarinet-sdk";
 import { join, resolve } from "path";
 import fs, { existsSync, readFileSync, rmSync } from "fs";
 import { createIsolatedTestEnvironment } from "./test.utils";
+import * as toml from "@iarna/toml";
```
moodmosaic: Are we sure `@iarna/toml` is isomorphic to the current `toml` handling/parsing?

citizen.ts
```diff
@@ -1,283 +1,138 @@
...
+    const rvContractsParentDir = join(manifestDir, ".rv-contracts");
+    try {
+      if (existsSync(rvContractsParentDir)) {
+        const contents = readdirSync(rvContractsParentDir);
+        if (contents.length === 0) {
+          rmSync(rvContractsParentDir, { recursive: true, force: true });
+        }
+      }
+    } catch {
+      // Ignore errors when cleaning up parent directory.
```
moodmosaic: I think this needs some thinking. Users won't know if cleanup failed, leading to disk bloat.

At the very least:

```ts
} catch (error) {
  radio.emit("logMessage", 
    `Warning: Failed to cleanup temporary directory: ${error.message}`
  );
}
```

.github/workflows/ci.yml
```diff
@@ -10,11 +10,11 @@ on:
 jobs:
   tests:
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 3
+    timeout-minutes: 10
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
```
moodmosaic: Why remove `macos-latest`? Blind spot for platform-specific issues.

citizen.ts
```diff
@@ -1,283 +1,138 @@
...
- * @param manifestPath The absolute path to the manifest file.
- * @param remoteDataSettings The remote data settings.
+ * @param manifestPath The path to the manifest file.
  * @param sutContractName The target contract name.
- * @returns The initialized simnet instance with all contracts deployed, with
- * the test contract treated as a first-class citizen of the target contract.
+ * @param radio The event emitter to send log messages to.
+ * @returns The initialized simnet.
  */
 export const issueFirstClassCitizenship = async (
```
moodmosaic: Instead of temporary files, what if something like this could work?

```ts
export const issueFirstClassCitizenship = async (
  manifestDir: string,
  manifestPath: string,
  sutContractName: string,
  radio: EventEmitter
): Promise<Simnet> => {
  // Parse once.
  const manifest = parseManifest(manifestPath);
  const deploymentPlan = await generateDeploymentPlan(manifest);
  
  // Build rendezvous in memory.
  const rendezvousContract = buildRendezvousContract(
    deploymentPlan, 
    sutContractName
  );
  
  // Init simnet with custom contract source resolver.
  const simnet = await initSimnetWithCustomContracts(
    manifest,
    { [sutContractName]: rendezvousContract }
  );
  
  return simnet;
};
```

*If* this works, it avoids temporary files, double initialization, disk I/O, cleanup complexity.

citizen.tests.ts
```diff
@@ -1,21 +1,18 @@
 import fc from "fast-check";
 import {
   buildRendezvousData,
-  deployContracts,
-  getContractSource,
-  getSbtcBalancesFromSimnet,
```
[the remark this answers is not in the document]
BowTiedRadone: Same as the previous comment, `initEmptySession` introduced the need to store the initial balances and restore, while `initSimnet` does that automatically.

citizen.ts
BowTiedRadone: With the latest approach (temporary isolated Clarinet project under the temporary file system), I believe `finally` is our best effort. Further handling/startup cleanup may unnecessary increase the complexity, especially since this is a short-term approach. Please feel free to resolve this comment if you agree 🙏

citizen.ts
```diff
@@ -1,283 +1,138 @@
...
-    const contractSource = rendezvousSourcesMap.get(contractId);
-    if (!contractSource) {
-      throw new Error(`Contract source not found for ${contractName}`);
-    }
-    return contractSource;
-  } else {
-    return readFileSync(join(manifestDir, contractProps.path), {
-      encoding: "utf-8",
-    });
+const initSimnetSilently = async (manifestPath: string): Promise<Simnet> => {
```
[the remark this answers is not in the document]
hugo-stacks: > I don't think there's a publicly exposed simnet method to only generate the deployment plan, but @hugo-stacks can correct me if I'm wrong

@BowTiedRadone The method exists but isn't exposed in JS. We could easily add it. 

~Maybe you can call `$ clarinet deployment generate --simnet` in a child process as a work
around~
⬆️ Actually this is a bad idea, the command and the SDK don't do _exactly_ the same thing

BowTiedRadone:
As soon as `3.12.0` clarinet release is out, this PR will be updated to use the newly exported `generateDeploymentPlan` method instead of initializing `simnet` twice. That's the only pending item at the moment.

BowTiedRadone:
Many thanks to @hugo-stacks for the latest https://github.com/stx-labs/clarinet release! Rendezvous is now faster and more reliable (as well as the CI). This PR was updated to use the latest, as well as the newly exported method that generates the deployment plan.

citizen.ts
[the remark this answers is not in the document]
BowTiedRadone: Many thanks @hugo-stacks! 🙏

wileyj (approved):

PR #215, opened 2025-12-31
Add failure persistence support for regression testing

### LINKED ISSUE #130: Support failure persistence via seed replay
Right now, failures are lost unless manually tracked. Rendezvous should persist failing test cases across runs. It should save the seed of a failing case, and replay it before generating new cases.

Expected behavior:  
- Failing cases are saved under `rendezvous-regressions/`.  
- Replayed first on next run.  
- (Optional?) Configurable path and grouping (e.g., by module or test).

BowTiedRadone:
This PR implements failure persistence and regression testing for Rendezvous. When tests fail, the failing seeds and configuration used are now automatically saved to `.rendezvous-regressions/`, allowing users to replay them on subsequent runs to prevent regressions.

**Key additions**
- New `--regr` CLI flag to control test execution: if present, run only the regression tests
- Persistent failure storage grouped by test type (invariant/property) with automatic replay
- Failures sorted by timestamp for chronological tracking

**Sample failure file**
`.rendezvous-regressions/ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.counter.json`:
```json
{
  "invariant": [
    {
      "seed": 901717247,
      "dial": "example/sip010.cjs",
      "numRuns": 15,
      "timestamp": 1767886534833
    },
    {
      "seed": -1374686468,
      "numRuns": 9,
      "timestamp": 1767886531457
    },
    {
      "seed": 1298457354,
      "dial": "example/sip010.cjs",
      "numRuns": 20,
      "timestamp": 1767883389583
    },
  ],
  "test": [
    {
      "seed": 1656313995,
      "numRuns": 6,
      "timestamp": 1767886553125
    },
    {
      "seed": 64830639,
      "numRuns": 11,
      "timestamp": 1767886546477
    },
    {
      "seed": 1593583466,
      "numRuns": 3,
      "timestamp": 1767886542907
    },
  ]
}
```

This brings Rendezvous closer to production-grade fuzzing by ensuring discovered bugs stay caught, while improving internal code organization and maintainability.

Closes #130.

### DIFF
--- a/app.tests.ts
+++ b/app.tests.ts
@@ -4,8 +4,8 @@ import { version } from "./package.json";
 import { resolve } from "path";
 import fs, { rmSync } from "fs";
 import { createIsolatedTestEnvironment } from "./test.utils";
-import { getFailureFilePath } from "./persistence";
 import { LOG_DIVIDER } from "./shared";
+import { getFailureFilePath } from "./persistence";
 
 const isolatedTestEnvPrefix = "rendezvous-test-app-";
 
@@ -214,7 +214,7 @@ describe("Command-line arguments handling", () => {
         `Using manifest path: ${manifestDirPlaceholder}/Clarinet.toml`,
         `Target contract: counter`,
         `Running regression tests.`,
-        `Regressions loaded from: ${resolve(getFailureFilePath("counter"))}`,
+        `Regressions loaded from: ${resolve(getFailureFilePath("ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.counter"))}`,
         LOG_DIVIDER,
       ],
     ],
@@ -282,7 +282,7 @@ describe("Command-line arguments handling", () => {
         `Using manifest path: ${manifestDirPlaceholder}/Clarinet.toml`,
         `Target contract: counter`,
         `Running regression tests.`,
-        `Regressions loaded from: ${resolve(getFailureFilePath("counter"))}`,
+        `Regressions loaded from: ${resolve(getFailureFilePath("ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.counter"))}`,
         LOG_DIVIDER,
       ],
     ],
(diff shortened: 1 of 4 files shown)


app.tests.ts
```diff
@@ -219,38 +246,54 @@ describe("Command-line arguments handling", () => {
...
       ],
     ],
     [
       ["manifest path", "contract name", "type=tESt (case-insensitive)"],
       ["node", "app.js", manifestDirPlaceholder, "counter", "tESt"],
       [
+        `-------------------------------------------------------------------------------`,
+        `Using manifest path: ${manifestDirPlaceholder}/Clarinet.toml`,
+        `Target contract: counter`,
+        `-------------------------------------------------------------------------------`,
```
wileyj: definitely nit-picky - but if this dash line is being copied/pasted many times for output, wouldn't it make sense to have it as a variable?

wileyj: maybe it's time to think about some form of standardized logger (or customizing an existing one?)

wileyj: full disclosure: i did something like this once, and went a bit overboard. when i was refactoring, it was a nightmare to replace all the `*********************` lines with the new separator 😆

property.ts
```diff
@@ -220,15 +181,157 @@ export const checkProperties = (
...
+      );
+
+      await propertyTest({
+        simnet,
+        targetContractName,
+        testContractId,
+        // If the number of runs that failed is less than 100, set it to the
+        // default value of 100. If more runs were needed to reproduce the
+        // failure, use the number of runs that failed.
+        runs: regression.numRuns < 100 ? 100 : regression.numRuns,
```
moodmosaic: If a failure occurred at run 50, why force 100 runs? This could mask issues where failures only occur early in the sequence.

BowTiedRadone: This is an extra safety mechanism. In practice, it means a historical failure might have failed after 5 runs, but the same sequence of events (seed) can be considered passing if, for a given user configuration (seed, dial, etc.), it passes the default number of runs used by Rendezvous. However, you raised a great point; this behavior should probably be documented more clearly/explicitly.

PR #248, opened 2026-04-14
Add support for config file and refactor run config parsing

### LINKED ISSUE #183: Allow property tests with specific wallet
I am using remote data because some wallet have a certain status. 

I would like to run tests with these specific wallets.

Concrete example: https://github.com/friedger/clarity-ccip-026

Some wallets hold mia tokens v1 or v2. These wallets should be used to test redemption of mia tokens.

BowTiedRadone:
This PR:
- decouples run configuration parsing from `main` to keep the entry point clean
- adds a `--config` flag for file-based configuration
- adds support for pre-specifying custom accounts (allowing mainnet addresses to be used without providing their private keys by leveraging simnet remote data) using the config file

Fixes #183.

### DIFF
--- a/app.ts
+++ b/app.ts
@@ -6,7 +6,7 @@ import { join, resolve } from "node:path";
 import { initSimnet } from "@stacks/clarinet-sdk";
 import { red } from "ansicolor";
 
-import { helpMessage, logRunConfig, parseCli, RunConfig } from "./cli";
+import { helpMessage, logRunConfig, parseCli, type RunConfig } from "./cli";
 import { resolveAccounts } from "./config";
 import { checkInvariants } from "./invariant";
 import { checkProperties } from "./property";
--- a/app.ts
+++ b/app.ts
@@ -2,19 +2,18 @@
 import { EventEmitter } from "node:events";
 import { existsSync } from "node:fs";
 import { join, resolve } from "node:path";
-import { parseArgs } from "node:util";
 
 import { initSimnet } from "@stacks/clarinet-sdk";
 import { red } from "ansicolor";
 
+import { helpMessage, logRunConfig, parseCli } from "./cli";
+import { resolveAccounts } from "./config";
 import { checkInvariants } from "./invariant";
-import { version } from "./package.json";
 import { checkProperties } from "./property";
 import {
   getContractNameFromContractId,
   getFunctionsFromContractInterfaces,
   getSimnetDeployerContractsInterfaces,
-  LOG_DIVIDER,
 } from "./shared";
 
 const logger = (log: string, logLevel: "log" | "error" | "info" = "log") => {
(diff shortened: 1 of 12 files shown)


cli.ts
```diff
@@ -0,0 +1,179 @@
...
+      accountsMode: fileConfig.accounts_mode ?? "overwrite",
+      configPath: options.config,
+    };
+  }
+
+  return {
+    manifestDir,
+    sutContractName,
+    type: normalizedType as "test" | "invariant",
+    seed: options.seed ? parseInt(options.seed, 10) : undefined,
```
moodmosaic: If a user passes `--seed=abc`, `parseInt("abc", 10)` produces `NaN`, which will silently propagate into fast-check as the seed. The config-file path validates this properly (`"seed" must be an integer`), so there's an asymmetry.

config.ts
```diff
@@ -0,0 +1,200 @@
...
+  const merged = new Map(
+    [...simnetAccounts].filter(([key]) => key !== "faucet"),
+  );
+
+  for (const [name, address] of configAccountsMap) {
+    merged.set(name, address);
+  }
+
+  // All addresses includes simnet addresses plus any new config addresses.
+  const allAddresses = [...merged.values()];
```
moodmosaic, line 213: This now excluded faucet, right? If this is intentional, a comment would help(?)

app.ts
```diff
@@ -44,143 +43,46 @@ export const getManifestFileName = (
...
-    --seed=<n>    Seed for replay functionality
-    --runs=<n>    Number of test iterations [default: 100]
-    --dial=<f>    Path to custom dialers file
-    --regr        Run regression tests only
-    --bail        Stop on first failure
-    -h, --help    Show this message
-
-  Learn more: https://stacks-network.github.io/rendezvous/
-  `;
+const parseCliOrExit = (argv: string[], radio: EventEmitter) => {
```
moodmosaic: The `catch` branch always calls `process.exit(1)` (which returns `never`). An explicit annotation would make the contract clearer for readers

```suggestion
const parseCliOrExit = (argv: string[], radio: EventEmitter): RunConfig | undefined => {
```

(Might have to be formatted to fit the 79-char width.)

moodmosaic (commented):
Very nice refactoring 👍 A few comments below (and inline):

---

I think, a user typing:
```
rv ./project counter test --config=rv.config.json --seed=999
```

would have no indication that `--seed=999` was discarded. Perhaps worth considering emitting a warning via the `radio` when *both* sources are present?

---

I also think that a typo like `{ "sedd": 42 }` would be silently accepted as an empty config(?) If that's the case, consider warning on unrecognized keys.

app.ts
BowTiedRadone: [5905ad8](https://github.com/stacks-network/rendezvous/pull/248/commits/5905ad843bf2395ca9a26369b57183060fb98ccd)

config.ts
BowTiedRadone: [8f74d90](https://github.com/stacks-network/rendezvous/pull/248/commits/8f74d90fddfed35207804548c7d83625e5a46f9e)

cli.ts
BowTiedRadone: [122f9be](https://github.com/stacks-network/rendezvous/pull/248/commits/122f9bef39859641cb19dbe5a2a58e8160b8c3e8)

BowTiedRadone:
@moodmosaic Thanks for the review. All should be addressed now

BowTiedRadone:
@moodmosaic Addressed the comments, this should be ready.

moodmosaic (approved):
🚢

friedger (commented):
Looks good. I didn't try it though.

PR #255, opened 2026-04-23
fix: clarify use of env simnet

friedger:
This PR
* adds a comment in clarity code in the README file that clarifies the use of `[#env(simnet)]`


BowTiedRadone (approved):
LGTM!

PR #258, opened 2026-04-30
Update all Rendezvous-specific functions access to private

### LINKED ISSUE #257: Use `define-private` for invariants, property tests, and discard functions
**Context**

Today Rendezvous expects invariants, property tests, and discard functions as `define-public` / `define-read-only`. If `;; #[env(simnet)]` annotation is missing, they can get shipped to mainnet and anyone can call them.

At a first sight this should work, since `clarinet-sdk` has `callPrivateFn`. `getContractsInterfaces()` already returns private functions with `access: "private"`.

**Required Changes**

- Filters: accept `access === "private"` in `invariant.ts:776`, `property.ts:583`, `property.ts:137`.
- Calls: switch to `callPrivateFn` at `invariant.ts:567`, `property.ts:412`, `property.ts:612`.
- `update-context`: flip to `define-private`; update calls at `invariant.ts:473`, `invariant.ts:727`.
- Examples: flip `invariant-*` / `test-*` / `can-*` / `update-context` in `example/contracts/*.clar`.
- Docs: recommend `define-private` in `README.md`, `docs/chapter_6.md`, `docs/chapter_7.md`.

**Done when:**

- Private invariants, tests, discard functions, and `update-context` run end-to-end.
- Examples and docs use `define-private`.
- Tests cover private discovery and invocation.

Thanks to @hugo-stacks for bringing this up.

BowTiedRadone:
This PR resolves #257.

**Changes**

- Invariants (`invariant-*`), property tests (`test-*`), discard functions (`can-*`), and `update-context` are now `define-private`. The harness invokes them via `simnet.callPrivateFn` (already exposed by `@stacks/clarinet-sdk`).
- New `RV_ACCESS` constant in `shared.ts` is the single source of truth for the access modifier per role; filters and test fixtures consume it instead of hardcoding `"private"`.
- All eight example contracts flipped.
- README and docs (chapters 4, 6, 7, 8) updated: code samples, prose conventions, and quoted CLI output (`(public)` → `(private)`).
- `filterSutFunctions` simplified: trusts the access filter alone.

**Breaking**

Contracts using the old conventions (e.g. `define-public` tests) will silently produce "no functions found".


docs/chapter_4.md
```diff
@@ -90,7 +90,7 @@ Here's an example of an invariant test that ensures a counter remains positive w
 This test demonstrates the key conventions for invariant tests in Rendezvous:
 
 - Function name starts with `invariant-`
-- Function is declared as `read-only` (not public)
+- Function is declared as `private`
```
moodmosaic: That one too?

property.ts
```diff
@@ -580,7 +582,7 @@ const filterTestFunctions = (
     Array.from(allFunctionsMap, ([contractId, functions]) => [
       contractId,
       functions.filter(
-        (f) => f.access === "public" && f.name.startsWith("test-"),
+        (f) => f.access === RV_ACCESS.test && f.name.startsWith("test-"),
```
moodmosaic: If the constant is accidentally changed, the test won’t catch it now. That is, the test became weaker.

shared.ts
```diff
@@ -31,6 +32,19 @@ import type { ImplementedTraitType, ImportedTraitType } from "./traits.types";
 export const LOG_DIVIDER =
   "-------------------------------------------------------------------------------";
 
+/**
+ * Single source of truth for the Clarity access modifier expected on each
+ * Rendezvous-managed function role (invariant / test / discard / context).
+ * Consumers should reference these instead of hardcoding the string so that a
+ * future change to the convention is a one-line edit.
+ */
+export const RV_ACCESS = {
```
moodmosaic: This provides very little benefit when code is covered by tests.

moodmosaic: What was the reason to declare this new structure? (Rule of three? DRY? Zen of Python?)

docs/chapter_4.md
friedger: Invariants are read-only by definition, aren't they? 
For developer experience, it might be easier to say all rv functions are private, for the logic invariants should be read-only

BowTiedRadone: @friedger We can turn invariants into read-only functions easily. However, the benefit of all the Rendezvous-related functions being private is UX simplicity (users only need to understand the return conventions, and not the access). What do you think?

BowTiedRadone: Actually, I'm going to change them back to read-only. They shouldn't be allowed to write, so read-only makes more sense indeed.

shared.ts
BowTiedRadone: Removed

property.ts
BowTiedRadone: Removed

docs/chapter_4.md
BowTiedRadone: [ee7f864](https://github.com/stacks-network/rendezvous/pull/258/commits/ee7f864180c8514d442e6619d0c89ccc95270267)

simone-stacks (commented):
On the thread that got invariants flipped back to `read-only`... on master the `can-*` discards were `define-read-only` and the harness called them via `callReadOnlyFn`. This PR moves them to `define-private` + `callPrivateFn`. But Friedger's argument that brought invariants back applies identically here: docs describe discards as predicates (`"returns true only if inputs are valid"`). Nothing in clarity stops a user now stuffing a state mutation into `can-test-foo` and silently polluting the test.

Either the read-only argument is right and `can-*` should follow invariants, or the invariant revert wasn't needed. I feel like mixing the two principles feels inconsistent, think it's worth flipping discards back to read-only too?

invariant.ts
```diff
@@ -724,7 +724,7 @@ export const initializeClarityContext = (
   functions: EnrichedContractInterfaceFunction[],
 ) => {
   functions.forEach((fn) => {
-    const { result: initialize } = simnet.callPublicFn(
```
simone-stacks, line 727: Small but important: if a user upgrades without flipping `update-context` to `define-private`, this `callPrivateFn` throws an opaque SDK error and the run dies with no hint about what to do.

Could we wrap with a contextual error here ("`update-context` must now be `define-private`...") or add a pre-flight check on the contract interface?

```suggestion
  functions.forEach((fn) => {
    let initialize;
    try {
      ({ result: initialize } = simnet.callPrivateFn(
        contractId,
        "update-context",
        [Cl.stringAscii(fn.name), Cl.uint(0)],
        simnet.deployer,
      ));
    } catch (e) {
      throw new Error(
        `Failed to call \`update-context\` for ${fn.name}. Since v1.x, ` +
        `\`update-context\` must be \`define-private\` (was \`define-public\`). ` +
        `Original error: ${e instanceof Error ? e.message : String(e)}`,
      );
    }
```

invariant.ts
simone-stacks, line 200: A migrating user with `(define-public invariant-foo)` (old convention) hits this exactly: `filterInvariantFunctions` now requires `read_only`, so the filter returns 0 and they land here with no hint about why. id append a one-liner pointing at the convention:

```suggestion
        `No invariant functions found for the "${targetContractName}" contract. Beware, for your contract may be exposed to unforeseen issues. (Note: as of v1.x, invariants must be \`define-read-only\`. See <link to migration doc>.)\n`,
```

simone-stacks, line 194: Less likely but still possible: if a user accidentally flips their actual SUT to `private` (or never had a real SUT and only Rendezvous-specific functions), this fires with no migration hint. I'd soften by pointing them at the convention check too:

```suggestion
        `No public functions found for the "${targetContractName}" contract. Without public functions, no state transitions can happen inside the contract, and the invariant test is not meaningful. (Note: as of v1.x, Rendezvous-specific functions \`test-*\` / \`can-*\` / \`update-context\` must be \`define-private\` — if you migrated your SUT too, flip it back to \`define-public\`.)\n`,

File: invariant.ts [hunk 5]
```diff
@@ -756,12 +756,7 @@ const filterSutFunctions = (
   new Map(
     Array.from(allFunctionsMap, ([contractId, functions]) => [
       contractId,
-      functions.filter(
-        (f) =>
-          f.access === "public" &&
```
simone-stacks, line 761: Dropping the `name !== "update-context" && !name.startsWith("test-")` guards is correct under the new convention but it removes the defensive net during migration. a contract with `(define-public test-foo)` now silently lands as SUT in invariant mode, `update-context` gets called against the test name, the context map gets polluted, and no warning.

Could we either keep a minimal name guard with a warning emit, or add a pre-flight sanity check that flags `public` functions named like `test-*`/`can-*`/`update-context`? cheap foot-gun preventer for the migration window.

example/contracts/slice.clar
simone-stacks, line 444: The comment block still says "Must be read-only", but `can-test-slice-list-int` below is `define-private`.

Since slice.clar is the canonical example file this is a direct contradiction in the file we tell people to copy from. Should be "Must be private." to match `chapter_8.md:543`. (or "Must be read-only.")

```suggestion
;; - Must be private.
```

example/contracts/reverse.clar
```diff
@@ -3,132 +3,122 @@
...
- (unwrap-panic
-  (as-max-len?
-   (concat (list item) seq)
-   u127))) ;; Replace with u4 to introduce a bug.
+    (item uint)
+    (seq (list 127 uint))
+  )
+  (unwrap-panic (as-max-len? (concat (list item) seq) u127))
+)
+;; Replace with u4 to introduce a bug.
```
simone-stacks: If I don't go wrong, this got orphaned by the .clar reformat bundled into commit 2fa74f3.

On master the `;; Replace with u4 to introduce a bug.` sat inline at the end of the `u127)))` line inside `reverse-redx-unsigned-list`.
After reformatting it now lives as a free-standing line between `reverse-redx-unsigned-list` and `reverse-list1`, and a reader can't tell which function it applies to.

Could you either restore inline placement, or rewrite to be unambiguous? something like:

```suggestion
;; In `reverse-redx-unsigned-list`, replace u127 with u4 to introduce a bug.
```

docs/chapter_6.md
```diff
@@ -659,9 +659,9 @@ This process allows Rendezvous to create meaningful state transitions and valida
 
 ### Example
 
-The `example` Clarinet project demonstrates this feature. The [send-tokens](https://github.com/stacks-network/rendezvous/blob/9c02aa7c2571b3795debc657bd433fd9bf7f19eb/example/contracts/send-tokens.clar) contract contains one public function and one property-based test that both accept trait references.
+The `example` Clarinet project demonstrates this feature. The [send-tokens](https://github.com/stacks-network/rendezvous/blob/12b3e4cc011c0029522b54e0e02342f9d47600eb/example/contracts/send-tokens.clar) contract contains one public function and one property-based test that both accept trait references.
```
simone-stacks: This permalink point to SHA 12b3e4c, where send-tokens.clar and rendezvous-token.clar are still `define-private` for invariants.
A reader following the link from "Example" sees code that contradicts the new convention in the surrounding prose. id bump the permalink SHA to a commit after ee7f864, or to master once this is merged

docs/chapter_6.md
```diff
@@ -659,9 +659,9 @@ This process allows Rendezvous to create meaningful state transitions and valida
...
 
-To enable testing, the project includes [rendezvous-token](https://github.com/stacks-network/rendezvous/blob/9c02aa7c2571b3795debc657bd433fd9bf7f19eb/example/contracts/rendezvous-token.clar), which implements the required trait.
+To enable testing, the project includes [rendezvous-token](https://github.com/stacks-network/rendezvous/blob/12b3e4cc011c0029522b54e0e02342f9d47600eb/example/contracts/rendezvous-token.clar), which implements the required trait.
```
simone-stacks: Same here

README.md
simone-stacks: `strategyFor` is described as "ready for use with `simnet.callPublicFn` or `simnet.callReadOnlyFn`", but the harness now uses `callPrivateFn` in 4 places (invariant.ts and property.ts)
A library consumer wanting to drive Rendezvous-style `test-*`/`can-*` through fast-check via the public API has no signal that `callPrivateFn` is the right entrypoint.

```suggestion
- `strategyFor(simnet, fn, allAddresses?, projectTraitImplementations?)` — returns an `fc.Arbitrary<ClarityValue[]>` ready for use with `simnet.callPublicFn`, `simnet.callReadOnlyFn`, or `simnet.callPrivateFn` (e.g. for Rendezvous-style private `test-*` / `can-*`). `allAddresses` restricts the principal pool (defaults to every account in the simnet); `projectTraitImplementations` reuses a precomputed trait map (defaults to extracting it from the simnet).
```

simone-stacks: I think the same line goes stale in two other files that aren't touched by this PR but describe the same harness behavior. What about patching them in the same go:

`lib.ts:78` (JSDoc, currently says "or similar"):

```ts
 * Generates a fast-check arbitrary that produces `ClarityValue[]` arrays
 * ready for use with `simnet.callPublicFn`, `simnet.callReadOnlyFn`, or
 * `simnet.callPrivateFn` (e.g. for Rendezvous-style private `test-*` / `can-*`).
```

`docs/chapter_9.md:36`:

```md
Returns a fast-check arbitrary that produces `ClarityValue[]` arrays — ready for use with `simnet.callPublicFn`, `simnet.callReadOnlyFn`, or `simnet.callPrivateFn` (e.g. for Rendezvous-style private `test-*` / `can-*`).
```

simone-stacks (commented):
Looks good, I just left some comments on what I think might need your attention

property.ts
simone-stacks, line 192: Same story on the property side: a user with `(define-public test-foo)` (old convention) hits this because `filterTestFunctions` now requires `private`. One-liner here too:

```suggestion
        `No test functions found for the "${targetContractName}" contract. (Note: as of v1.x, test functions must be \`define-private\`. See <link to migration doc>.)\n`,
```

invariant.ts
radu-stacks: @simone-stacks We'll make sure to include a breaking change message in the release notes. Given this will be shipped in `v1.0.0`, it makes perfect sense that breaking changes exist.

radu-stacks: Also, the SDK message is clean enough when Rendezvous tries to call a private function that is not found.

invariant.ts
radu-stacks: Same comment regarding the breaking change. Totally acceptable. Including a `1.x`-specific message introduces redundancy (think about the changes needed in the future when switching to `2.x` for example)

radu-stacks: Basically targeting a contract without SUT function with invariant testing doesn't make sense (since the contract cannot be treated as a state machine). No extra note is needed.

docs/chapter_6.md
radu-stacks: Great catch!

property.ts
radu-stacks: Acceptable. No extra note needed.

invariant.ts
radu-stacks: Acceptable for `1.0.0`.

README.md
radu-stacks: [5c540e6](https://github.com/stacks-network/rendezvous/pull/258/commits/5c540e6109a56cbd81b4d999105524ebc6a99ee1)

example/contracts/slice.clar
radu-stacks: Thanks for catching this. Now that I revisit, it makes more sense for discard functions to be read-only (shouldn't be able to write and modify the contract state).

radu-stacks: [9a6a8f5](https://github.com/stacks-network/rendezvous/pull/258/commits/9a6a8f59455263be3142d5ad5b7b9bff0c900796)

example/contracts/reverse.clar
radu-stacks: [1e9fe78](https://github.com/stacks-network/rendezvous/pull/258/commits/1e9fe78b07d8cd971843634f3915186295efb414)

simone-stacks (approved):
lgtm!

PR #259, opened 2026-05-12
Add trophy case and submit first vulnerability

BowTiedRadone:
This PR adds a trophy case to `README`, links docs to it in `SUMMARY`, and records the first vulnerability reported via DM.

Congrats to @Rapha-btc!

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,15 @@ fc.assert(
 
 ---
 
+### Trophy Case
+
+Real bugs and security vulnerabilities uncovered with `rv`. To add a finding, open a pull request appending a row to the table below.
+
+| Date | Project | Vulnerability | Invariant/Property | Version | Submitter |
+| ---- | ------- | ------------- | ------------------ | ------- | --------- |
+
+---
+
 ### Documentation
 
 For full documentation, see the official [Rendezvous Book](https://stacks-network.github.io/rendezvous/).
--- a/README.md
+++ b/README.md
@@ -145,8 +145,9 @@ fc.assert(
 
 Real bugs and security vulnerabilities uncovered with `rv`. To add a finding, open a pull request appending a row to the table below.
 
-| Date | Project | Vulnerability | Invariant/Property | Version | Submitter |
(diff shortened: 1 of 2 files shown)


moodmosaic (commented):
Happy to see this!

Rapha-btc:
Thanks for kicking off the trophy case, @BowTiedRadone — humbled to be the first entry 🙏

Quick context on what RV caught:

While preparing Jing V3, I ran @stacks/rendezvous 500-run sweeps with 13 state invariants against both markets (**sBTC/STX** + **sBTC/USDCx**). 12 of them checked list/totals/map consistency the "obvious" way. The 13th — `invariant-balance-eq-cycle-totals` — compared the contract's actual token balance against the sum of declared cycle-totals across all cycles.

That 13th invariant caught a **cancel-cycle × small-share-roll** bug:

When `close-deposits` ran `small-share-filter`, it rolled sub-threshold depositors' lists + totals forward to next-cycle. If `cancel-cycle` then fired on the current cycle, our roll code was **overwriting** next-cycle's totals instead of **merging** — the totals from the small-share roll silently got wiped. List/totals/map stayed mutually consistent at the wrong values, so the other 12 invariants happily passed. Only the actual-balance-vs-declared-totals check saw the corruption.

Caught within ~30 random tx sequences after adding the invariant. Fix is a one-liner in `roll-depositor-lists` (concat next-cycle's existing list with cycle's, then `+` the totals). Verified by reverting the fix in `.build/` — RV trips the invariant deterministically. With the fix, both markets pass 500-run sweeps clean.

**v2 was paused as soon as the bug was identified; v3 ships with the fix plus other improvements** — so no exposure remained.

**Takeaway for other Clarity teams:** when fuzzing balance-preserving protocols, always add a **conservation invariant** (actual on-chain balance vs sum of declared accounting). Internal-consistency checks alone can miss the entire class of "the accounting is corrupted but consistently corrupted."

Full writeup with repro / failing invariant / fix details: https://github.com/Rapha-btc/jing-contracts-v3/blob/master/tests/rv/README.md

wileyj (approved):

PR #263, opened 2026-05-28
Post-transfer fixes

radu-stacks:
This PR fixes links after transfer to stx-labs. It also updates the docs publishing-related jobs to use the native GH Pages actions (previous jobs stopped deploying at some point in the past due to the lack of a Jekyll build).


jannik-stacks (approved):
looks good

PR #265, opened 2026-06-11
Add error handling for getContractAST in project traits

friedger:
This PR
* adds try/catch where getContractAST may throw due to non-introspectable contracts.
* fixes #264

### DIFF
--- a/traits.ts
+++ b/traits.ts
@@ -553,7 +553,17 @@ export const extractProjectTraitImplementations = (simnet: Simnet) => {
   const projectTraitImplementations = allProjectContracts.reduce<
     Record<string, ImplementedTraitType[]>
   >((acc, contractId) => {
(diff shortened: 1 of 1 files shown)


traits.ts
```diff
@@ -553,7 +553,17 @@ export const extractProjectTraitImplementations = (simnet: Simnet) => {
   const projectTraitImplementations = allProjectContracts.reduce<
     Record<string, ImplementedTraitType[]>
   >((acc, contractId) => {
-    const ast = simnet.getContractAST(contractId);
+    // Some deployed contracts (notably remote `requirements` contracts) have
```
hugo-stacks, line 556: Important note to avoid confusion;

Remote doesn't rely on `requirements` _at all_.
It simply fetch whatever data your request from the Node RPC enpoints.

`requirements` are still being used for local development

